PR #1, opened 2017-09-18
Create TwitterGuidelines.md

kelle:


eteq:
I stumbled across this and I think it's probably better to merge this instead of leaving it just hanging as a PR since it's been gathering dust for some time.  Can always be updated in the future.

PR #2, opened 2017-09-18
policy regarding responding to provocative comments

kelle:
added language about not responding if unable to respond in a helpful and supportive way.


eteq:
Haven't quite had a chance to look at this, but one clarification: is the idea that this be the canonical copy of the code of conduct?  The current "official" one is the one on the web site - I have a feeling these will get out of sync if there are multiple copies of it.

code-of-conduct.md
```diff
@@ -7,6 +7,9 @@ As members of the community,
...
 - We pledge to welcome questions and answer them respectfully, paying particular attention to those new to the community. We pledge to provide respectful criticisms and feedback in forums, especially in discussion threads resulting from code contributions.
 - We pledge to be conscientious of the perceptions of the wider community and to respond to criticism respectfully. We will strive to model behaviors that encourage productive debate and disagreement, both within our community and where we are criticized. We will treat those outside our community with the same respect as people within our community.
+
+*If any community member ever feels like they can not respond to a public forum or private digital communication regarding the Astropy Project or Python in a supportive and helpful way, do not respond. Please forward the communication to a [Community Coordinator](http://www.astropy.org/team.html), the Coordination Committee (coordinators@astropy.org) or confidential@astropy.org.*
```
eteq: Hmm, I don't like the way this renders.  It breaks up the "We pledge" flow below.  I wonder if it might be better if it actually be put directly into the web page copy of the CoC as a footnote or similar sort of form that doesn't break up the flow of the primary bullet points?

eteq: (I guess the solution to this depends somewhat on my comment further up about whether this is the "true" version of the CoC or if this is just a draft copy for us to integrate into the web site as the "true" version.)

kelle:
I'll have to think about the substantive comments and respond later.

In response to the question about the CoC document living in this repo, honestly, it just felt wrong to open this discussion in the repo about the webpage. Yes, this is the place where we can hash out the language but the "live" CoC will be the one on the webpage. I think that the CoC will change infrequently enough that it would be too hard to keep the webpage in sync with this document.

code-of-conduct.md
astrofrog: Yes I was actually thinking about this again and I think this is more of a bit about how to act on the CC than a new bullet point. It could be mentioned around the same place where we mention who to contact

taldcroft: Can we just pick one point of contact for sending the communication instead of three choices?  I would just say to send to the coordination committee (coordinators@astropy.org).  Where does confidential@astropy.org go to?

kelle: It's important to give people options for who to contact, especially since our community is so small. For example, what if the problem is with someone on the coordinating committee?

Confidential@astropy.org will go to the Ombudsperson (that should be added!).

kelle: I modified it and moved the language to the end so as to not break up the flow. I also added explicitly that confidential goes to the Ombudsperson.

code-of-conduct.md
```diff
@@ -7,10 +7,12 @@ As members of the community,
...
 - We pledge to be conscientious of the perceptions of the wider community and to respond to criticism respectfully. We will strive to model behaviors that encourage productive debate and disagreement, both within our community and where we are criticized. We will treat those outside our community with the same respect as people within our community.
-- We pledge to help the entire community follow the code of conduct, and to not remain silent when we see violations of the code of conduct. We will take action when members of our community violate this code such as contacting confidential@astropy.org (all emails sent to this address will be treated with the strictest confidence) or talking privately with the person.
+- We pledge to help the entire community follow the code of conduct, and to not remain silent when we see violations of the code of conduct. We will take action when members of our community violate this code. such as contacting the Ombudsperson at confidential@astropy.org (all emails sent to this address will be treated with the strictest confidence) or talking privately with the person.
```
taldcroft: Perhaps start with `talking privately with the person` and follow with sending email to the Ombudsperson?  That also makes the sentence easier to parse since there is this long parenthetical note.  That note can be shortened to `(emails to this adress are treated with strictest confidence)`.

eteq:
@kelle - one clarification: do you intend for the final version of this to have the new text be a footnote or otherwise attached to the bullet point it was connected to before?  I think it's still a bit jarring to have this sentence in there because all of the rest is positive and affirming ("we pledge"), while the added bit is a "don't do this". It also right now sort of duplicates the bullet point on " respond to criticism respectfully".  I think it works great as a footnote or otherwise "attached" element to that bullet, as it makes it clear what to do, but standing on it's own like it is in this document, it's a bit jarring.

kelle:
No, I was not envisioning it as a footnote. I just moved it to the bottom and outside of the bulleted list. To me, it seems fine there. 

I think we could just add this for now and consider the possibility of including more examples of Do's and  Don'ts in a future iteration. The more I work on these CoC type issues, the more I see the need for some explicit examples. Saying "be respectful" is really not at all useful and means different things to different people.

kelle:
I am gonna say that we didn't reach consensus on this and close it.

PR #9, opened 2019-10-15
slight change to goal wording

kelle:
this change is from Gurtina Besla after my talk at Inclusive Astronomy.


eteq:
(I had to force-push to update this due to me re-organizing of the repo, but that should be fine)

PR #48, opened 2020-01-17
Draft email with clearification on how Moore grant can be spend

### LINKED ISSUE #47: Draft questions about grant rules
We have a number of questions about the administration of the grant, and it would be useful to clarify the rules:

Here are some examples:
What is the period?
Should this extend the period?  
Is there any rules about how the extra money has to be spent? `
How closely do we have to align to the original budget?

Proposal is to collaboratively draft the letter with a deadline of the next Finance committee meeting.

hamogu:
The finance comitee is looking at the Moore grant application. The application
lays out a budget, but it seems that the rules for spending are not as strict
as some of us are used to from e.g. NASA grants. However, its always better
to be upfront and ask early, rather than to find out later that the deadline for
spending money has already passed and money has to be returned.

Thus, we plan to contact the Moore foundation to ask for the rules.

**Deadline: Feb 15 2020**

This PR will be merged and the email send after the deadline, so any comments
have to be done before that date.

closes #47

### DIFF
--- a/finance/clearifying_question_to_moore.txt
+++ b/finance/clearifying_question_to_moore.txt
@@ -1,0 +1,32 @@
+Dear [INSERT CONTACT AT MOORE FOUNDATION],
+
+thank you so much for giving the grant "Sustaining and Growing the Astropy
+Project" to us. As you know from the grant application, we are still working
+out a formal government and management structure, because the astropy project
+has not received grants like this before.
+
+However, many astropy members are used to government issued contracts and
+grants (e.g. NSF, NASA) from their non-astropy science work, where every grant
+comes with a separate set of rules.
+
+We'd thus like to clarify a few points about timeline and spending about the
+Moore grant early on - in particular since you awarded us more money than we
+originally budgeted and we are trying to decide now, how that extra money will
+benefit the project most.
+
+- We proposed work over three years of time. When did the clock for that
+  formally start?  
+
+- We are keen to get the work done, but if we don't spend
+  the money as fast as we thought we would, could the grant period be extended,
+  e.g. by another year?  
+
+- For the application, we estimated a budget and broke that down into different
+  sub-categories. Obviously, the costs won't come in exactly like that to the
+  cent. Do you have a guideline for us like "if you shift more than XX k$ from
+  category A to category B, please check with us in advance"?
+
+
+Yours sincerely,
+
+[NAME HERE]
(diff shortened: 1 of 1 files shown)


finance/clearifying_question_to_moore.txt
```diff
@@ -0,0 +1,32 @@
+Dear [INSERT CONTACT AT MOORE FOUNDATION],
+
+thank you so much for giving the grant "Sustaining and Growing the Astropy
+Project" to us. As you know from the grant application, we are still working
+out a formal government and management structure, because the astropy project
```
jdswinbank: I think we should capitalize “Astropy Project” throughout.

finance/clearifying_question_to_moore.txt
```diff
@@ -0,0 +1,32 @@
...
+However, many astropy members are used to government issued contracts and
+grants (e.g. NSF, NASA) from their non-astropy science work, where every grant
+comes with a separate set of rules.
+
+We'd thus like to clarify a few points about timeline and spending about the
+Moore grant early on - in particular since you awarded us more money than we
+originally budgeted and we are trying to decide now, how that extra money will
+benefit the project most.
+
+- We proposed work over three years of time. When did the clock for that
```
jdswinbank: I'd state this as “What is the period of performance for the work performed under this grant?”, or similar — that avoids any potential issues with rounding or similar in “start date + three years”.

finance/clearifying_question_to_moore.txt
```diff
@@ -0,0 +1,32 @@
...
+  formally start?  
+
+- We are keen to get the work done, but if we don't spend
```
jdswinbank: Should probably clarify that this is a no-cost extension.

hamogu: I was trying to avoid wording that has a specific legal meaning in NASA/NSF context.

hamogu:
I'll just removed the "WIP" from the title. I meant to indicate that the interim finance committee is still working on it (as in "Moritz wrote it but the others have not seen it yet), so other people might save time by waiting until we agree on a draft and then commenting on the consensus draft,  but that's probably more complicated than necessary.

hamogu:
We also discussed overhead (called "indirect cost" by the Moore foundation), but that is described in quite some detail here: https://www.moore.org/docs/default-source/Grantee-Resources/indirect-cost-policy.pdf so I don't think there is anything to ask at this point.

In our specific case, essentially all money we plan to spend is excluded from the calculation of the indirect costs", so we need to check with NumFOCUS and might ask them to apply a fixed fee, not a percentage, to this grant.

jdswinbank:
Thanks for the changes, and also for the link to that Moore Foundation document! 👍

hamogu:
We'll see if any of these questions can be answered by carefully reading the contract again or if NumFOCUS knows answers for these because they've worked with Moore for other grants before.

hamogu:
The answers to these questions are all in the Moore grant contract, thus I'm closing this issue.

PR #57, opened 2020-02-23
Interaction guide from AAS 235

eblur:
I put together a list of recommendation for interacting with people at the NumFOCUS booth at AAS 235. In general, it provides my advice for setting a respectful tone in code learning / instructional environments.

### DIFF
--- a/drafts/interaction-guide.md
+++ b/drafts/interaction-guide.md
@@ -2,10 +2,11 @@
 **This list of recommendations was provided to volunteers for the 
 NumFOCUS booth at AAS 235.**
 
-The code of conduct sets the ground state for professionalism, but
-there are many ways to inadvertently cause harm during interactions
-with users. This short guide offers a few tips for communicating
-effectively and respectfully.
+The Astropy code of conduct (CoC) seeks to create a welcoming 
+environment for all. This short guide focuses on common missteps 
+in user interaction that can lead to a poor user experience. The
+example cases and their complementary responses are meant to train effective 
+and respectful communication that adheres to the spirit of the Astropy CoC.
 
 1. No code bashing, directly or indirectly. No scoffing, rolling your
 eyes, or questioning users' choices. Everyone chooses different tools
--- a/drafts/interaction-guide.md
+++ b/drafts/interaction-guide.md
@@ -1,33 +1,42 @@
 
+**This list of recommendations was provided to volunteers for the 
+NumFOCUS booth at AAS 235.**
+
 The code of conduct sets the ground state for professionalism, but
 there are many ways to inadvertently cause harm during interactions
 with users. This short guide offers a few tips for communicating
 effectively and respectfully.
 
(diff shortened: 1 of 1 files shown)


drafts/interaction-guide.md
```diff
@@ -0,0 +1,73 @@
+
+**This list of recommendations was provided to volunteers for the 
+NumFOCUS booth at AAS 235.**
+
+The code of conduct sets the ground state for professionalism, but
+there are many ways to inadvertently cause harm during interactions
```
hamogu: I was not at a AAS, but this wording seems to imply that it's really dangerous to be at NumFOCUS booth. I would not like to volunteer there, because there are so many ways I can cause harm. Maybe I better don't do anything!

To avoid that impression, I suggest to tone this down a little: Just deleting the second have of the sentence "but ..." would work.

kelle: Actually, @hamogu , I think that's not too far off from what we're getting at. We've had a couple of cases where one interaction has really hurt someone's feelings and turned them off from the Project entirely. We want to be sure people who are in "public-facing" positions understand how easy this is to do and to put significant effort into doing no harm. I think we just need to be more clear about what we mean by "harm". 

How about, "There are many ways to inadvertently insult or condescend to our users."

hamogu: If you want to be that explicit, because you experienced people at the booth doing real damage, then maybe it has to be the way you originally suggested.

I'm just saying, that, if you want to recruit me as a volunteer to run a booth, you'll have a hard time to convince me if it starts out with pointing out that there are so many ways I can harm the cause.

I'd prefer a more gentle wording in the written policy and a verbal explanation of what to look out for. Written policies get read without context - but maybe the bottom line is that booth like that should not be staffed by random people, but only by per-screened professionals who have proven that they handle the outreach efforts appropriately (but if that's that case, then such a policy would not be needed).

Either way, I was not there, I'm just reading the PR here and raising the flag "this policy would not want me to volunteer time". If you want to be gentle, don't include "there are many ways to insult", instead say "User may feel itimided or insulted easily, so ...". If you think that a *gentle* policy is not enough, go for your original text, but know that you are turning away some fraction of potential volunteers in the future.

eblur:
Hi folks, I recently re-read the CoC and decided to change the intro to this document. I note that  this guide doesn't go *beyond* the CoC, but is well within in the spirit of the CoC. 

Despite it being under the purview of the CoC, this set of common micro-aggressions (which I've softened to "missteps") are things I've noticed both inside and outside of the Astropy community. So it's still important to name them.

I think this also addresses @hamogu 's concern.

pllim:
That looks great, @eblur ! Does this address the issue you brought up in the governance meeting today? If not, please let me know. If there are ways to improve anything over at `astropy` documentation, they will be welcome too.

Thank you!

pllim:
This got approved 3 years ago and will only be in draft folder, so let's just merge it.

PR #90, opened 2020-05-10
Update affiliated package review process to delegate editorial work

astrofrog:
This updates the affiliated package review process to delegate editorial work to people in named roles:

https://github.com/astropy/project/issues/87

I will also open a PR to the website repo to add the new role (done: https://github.com/astropy/astropy.github.com/pull/376)

### DIFF
--- a/affiliated/affiliated_package_review_guidelines.md
+++ b/affiliated/affiliated_package_review_guidelines.md
@@ -78,7 +78,7 @@ ecosystem - does it make use of existing functionality, or does it duplicate it?
 <table>
 <tr>
 <td width=150><img src="https://img.shields.io/badge/Red-red.svg" alt="Red"></td>
-<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead.</td>
+<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead, or duplicates functionality found in Astropy or other affiliated packages.</td>
 </tr>
 <tr>
 <td><img src="https://img.shields.io/badge/Orange-orange.svg" alt="Orange"></td>
--- a/affiliated/affiliated_package_review_guidelines.md
+++ b/affiliated/affiliated_package_review_guidelines.md
@@ -78,7 +78,7 @@ ecosystem - does it make use of existing functionality, or does it duplicate it?
 <table>
 <tr>
 <td width=150><img src="https://img.shields.io/badge/Red-red.svg" alt="Red"></td>
-<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead, or duplicates functionality found in Astropy or other affiliated packages.</td>
+<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead, or unnecessarily duplicates functionality found in Astropy or other affiliated packages.</td>
 </tr>
 <tr>
 <td><img src="https://img.shields.io/badge/Orange-orange.svg" alt="Orange"></td>
--- a/affiliated/affiliated_package_review_process.md
+++ b/affiliated/affiliated_package_review_process.md
@@ -84,15 +84,15 @@ posted on the pull request and also in a reply to your email.
 
 *These instructions are for editors and reviewers.*
 
-The review process will be overseen and coordinated by individuals that are in
+The review process will be overseen and coordinated by the individuals that are in
 the *Affiliated package review editor* role. This role
 may be filled by one or more individuals. For the rest of this section, we
 use the singular term *editor*, although multiple people may share this role
-(and choose how to do this). The
+(and it is at their discretion to decide how to divide up work). The
 editor will start off by finding someone to carry out the main review of the
 package (hereafter the *reviewer*). In this sense the role of the editor is
-similar to that of a journal editor (and hence the role name). However this is
-only partially true, because here the editor may add their views to the review,
+similar to that of a journal editor (and hence the role name). However this analogy
+is only partially accurate, because here the editor may add their views to the review,
 unlike a journal editor who is often strictly impartial on the review itself.
 The pool of available reviewers will be anyone who has an official role on [the
 Astropy team](http://www.astropy.org/team.html) or anyone who has participated
@@ -108,7 +108,7 @@ document. Emails should only be sent to the next potential reviewer(s) - not
 many possible reviewers at the same time. The editor should let the potential
 reviewer know at this point that the editor may also be doing a review and
 synthesizing the results into a single review. Once a reviewer accepts, the
-editor send them an email with detailed review instructions (which includes a
+editor sends them an email with detailed review instructions (which includes a
 link to the `review guidelines
 <https://github.com/astropy/project/blob/master/affiliated/affiliated_package_review_guidelines.md>`_),
 using the template at the bottom of this document. The editor then leaves a
--- a/affiliated/affiliated_package_review_process.md
+++ b/affiliated/affiliated_package_review_process.md
@@ -102,12 +102,14 @@ expertise required, and if scientific or algorithmic expertise is needed which
 is not available in the project, reviewers can be requested from outside the
 project at the editor's discretion.
 
-The editor sends out an email to possible reviewers, individually, to find
-someone who is willing to review the package, using the template at the bottom
-of this document. They should let the potential reviewer know at this point that
-the editor may also be doing a review and synthesizing the results into a
-single review. Once a reviewer accepts, the editor send them an email with
-detailed review instructions (which includes a link to the `review guidelines
+The editor sends out an email to possible reviewers to find someone who is
+willing to review the package, using the template at the bottom of this
+document. Emails should only be sent to the next potential reviewer(s) - not
+many possible reviewers at the same time. The editor should let the potential
+reviewer know at this point that the editor may also be doing a review and
+synthesizing the results into a single review. Once a reviewer accepts, the
+editor send them an email with detailed review instructions (which includes a
+link to the `review guidelines
 <https://github.com/astropy/project/blob/master/affiliated/affiliated_package_review_guidelines.md>`_),
 using the template at the bottom of this document. The editor then leaves a
 message on the pull request proposing the affiliated package saying:
--- a/affiliated/affiliated_package_review_process.md
+++ b/affiliated/affiliated_package_review_process.md
@@ -2,7 +2,7 @@
 
 This document describes the procedure for proposing, reviewing, and making a
 decision about Astropy-affiliated packages. This is meant as a record of the
-procedure and not the first place people should go to for - for instructions
+procedure and not the first place people should go to - for instructions
 on proposing an affiliated package, see [**Becoming an Affiliated Package**](http://www.astropy.org/affiliated/index.html).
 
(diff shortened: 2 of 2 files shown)


affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,46 @@ posted on the pull request and also in a reply to your email.
...
 
 Once the reviewer has finished, they send their review back by email to the
-coordinator, who then does their own review and then writes up a review using
-the review template which includes a table with badges (template at the bottom
-of this document). They then share it with the other coordination committee
-members to check for consensus. The review including the final decision (accept
-or reject) is then posted on the original pull request. If the package is
+editor, who then does their own review if needed and then writes up a review
+using the review template which includes a table with badges (template at the
+bottom of this document). At this point, the editors make a decision to access
```
pllim, line 117: Typo:
```suggestion
bottom of this document). At this point, the editors make a decision to accept
```

Maybe link to https://github.com/astropy/project/blob/master/affiliated/affiliated_package_review_guidelines.md in this section too.

affiliated/affiliated_package_review_process.md
```diff
@@ -179,7 +184,7 @@ You can find our review guidelines and instructions at the following address:
   https://github.com/astropy/astropy-procedures/blob/master/documents/affiliated_package_review_guidelines.md
```
pllim, line 184:
```suggestion
  https://github.com/astropy/project/blob/master/affiliated/affiliated_package_review_guidelines.md
```

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,46 @@ posted on the pull request and also in a reply to your email.
...
+package (hereafter the *reviewer*). In this sense the role of the editor is
+similar to that of a journal editor (and hence the role name). However this is
+only partially true, because the editor may add their views to the review,
+unlike a journal editor who is often strictly impartial on the review itself.
+The pool of available reviewers will be anyone who has an official role on [the
+Astropy team](http://www.astropy.org/team.html) or anyone who has participated
+in the project enough to be familiar with the project guidelines and requirements.
+In some cases, it may be necessary to have multiple reviewers depending on the
+expertise required, and if scientific or algorithmic expertise is needed which
+is not available in the project, reviewers can be requested from outside the
```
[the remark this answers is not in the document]
astrofrog: editors, fixed

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,46 @@ posted on the pull request and also in a reply to your email.
...
+project.
+
+The editor sends out an email to possible reviewers one at a time to find
```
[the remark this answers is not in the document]
astrofrog: The one at a time meant that if you need e.g. one reviewer, not to email five and then pick the first to reply. But that's perhaps too much detail. I've changed this to 'individually' to make it clear not to send blanket emails to 5 people at the same time.

How long the editors wait for reviewers to reply is up to them.

The reviewers are anonymous unless they say they are happy to not be. However, I think @eteq and I discussed this and thought it would make sense to have it the other way around - not anonymous unless requested by the reviewer. I will make that update in a separate PR though.

affiliated/affiliated_package_review_process.md
astrofrog: Done

astrofrog:
@pllim - thanks for the detailed review!

> Usage of "editor" and "editors" are inconsistent, which makes it a little confusing.

Fixed

> What if this package that is asking for review contains same functionality of another already accepted affiliated package? Will accepting both go against the "improving reuse" mantra?

This would go against the 'ecosystem integration' criterion in the guidelines, which I've now tried to clarify further. Basically we don't want to allow duplicate packages without good reason.

> I think conflict of interest cases should be clarified here too. For example, editor/reviewer knows package author very well. Hopefully, people will be honest and declare such things themselves.

I added a sentence to the email to reviewers about this.

>the text on L5 does not quite make sense to me: not the first place people should go to for - for instructions... (Why is "for" repeated? seems like an incomplete sentence before the dash.)

Fixed

> Add it at the top of the registry.json file -- I prefer alphabetical order by package name, but maybe that's just me. grimacing

I agree that would be nice, but first we then need to make sure it's actually sorted in the registry.json. Feel free to do this and then open a follow-up PR once this one is merged?


> If the package is rejected, the decision is posted to the pull request, and the pull request is closed. -- Is there a grace period before the same package can re-apply?

Not at the moment, but we could choose to add this as a guideline - but maybe in a separate PR?

> Name: Description: Repository: Documentation: -- Line break did not render in "Template email with review request"

Fixed

> Dear/hi should just be "Hi," otherwise straight copy-paste without removing one of them results in awkward email greeting.

Changed it to 'dear' as default and editors can choose to change to Hi

> Badge colors in review guidelines -- Why orange and not yellow; might as well follow the actual traffic light color scheme that it has mentioned? (See #92) Actually... I think that document needs some edits too to be consistent with the changes here.

Let's see if we can agree on this after this PR is merged?

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,51 @@ posted on the pull request and also in a reply to your email.
...
+unlike a journal editor who is often strictly impartial on the review itself.
+The pool of available reviewers will be anyone who has an official role on [the
+Astropy team](http://www.astropy.org/team.html) or anyone who has participated
+in the Astropy project enough to be familiar with the project guidelines and requirements.
+In some cases, it may be necessary to have multiple reviewers depending on the
+expertise required, and if scientific or algorithmic expertise is needed which
+is not available in the project, reviewers can be requested from outside the
+project at the editor's discretion.
+
+The editor sends out an email to possible reviewers, individually, to find
```
pllim, line 105: Hmm, "individually" doesn't really explain to me that you disallow blanket emails to 5 different reviewers when you only will pick one of them. I think this expectation should be its own full sentence.

astrofrog: Does the latest commit clarify things? If not, can you make a suggestion?

astrofrog:
> I am not sure... What if the package author harasses the reviewer upon unfavorable review or something?

Ok yeah let's keep it as-is for now and if people feel strongly about the non-anonymous by default then we can open a separate PR/issue and discuss there.

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,53 @@ posted on the pull request and also in a reply to your email.
...
+in the Astropy project enough to be familiar with the project guidelines and requirements.
+In some cases, it may be necessary to have multiple reviewers depending on the
+expertise required, and if scientific or algorithmic expertise is needed which
+is not available in the project, reviewers can be requested from outside the
+project at the editor's discretion.
+
+The editor sends out an email to possible reviewers to find someone who is
+willing to review the package, using the template at the bottom of this
+document. Emails should only be sent to the next potential reviewer(s) - not
+many possible reviewers at the same time. The editor should let the potential
```
pllim, line 108: Something like this would be clearer for me but I don't know if that is too much. Also, you would need to break that long sentence up if you decide to adapt it here. Thanks!

```suggestion
many possible reviewers at the same time; For instance, if the review needs one reviewer, a request should first be sent to Reviewer A only and then to Reviewer B after an X amount of time has elapsed without any reply from Reviewer A (recommended X is 4 weeks and an email telling Reviewer A that their service is no longer needed should also be sent out too). The editor should let the potential
```

astrofrog: Thanks! I think this is getting too specific though - I think the editors should decide on such policies internally. Also in the above scenario I think you'd want to sent reminders to reviewers etc and that really is too much detail here. My point here originally was more that the editor shouldn't just email all core maintainers asking for a volunteer, but contact people individually.

affiliated/affiliated_package_review_guidelines.md
```diff
@@ -78,7 +78,7 @@ ecosystem - does it make use of existing functionality, or does it duplicate it?
 <table>
 <tr>
 <td width=150><img src="https://img.shields.io/badge/Red-red.svg" alt="Red"></td>
-<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead.</td>
+<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead, or duplicates functionality found in Astropy or other affiliated packages.</td>
```
eteq, line 81:
```suggestion
<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead, or unnecessarily duplicates functionality found in Astropy or other affiliated packages.</td>
```

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,53 @@ posted on the pull request and also in a reply to your email.
...
-The coordinator sends out an email to possible reviewers one at a time to find
-someone who is willing to review the package, using the template at the bottom
-of this document. Let them know at this point that the coordinator will also be
-doing a review and synthesizing the results into a single review. Once a
-reviewer accepts, the coordinator sends them an email with detailed review
-instructions, using the template at the bottom of this document. The coordinator
-then leaves a message on the pull request saying:
+*These instructions are for editors and reviewers.*
+
+The review process will be overseen and coordinated by individuals that are in
```
eteq, line 87:
```suggestion
The review process will be overseen and coordinated by the individuals that are in
```

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,53 @@ posted on the pull request and also in a reply to your email.
...
+the *Affiliated package review editor* role. This role
+may be filled by one or more individuals. For the rest of this section, we
+use the singular term *editor*, although multiple people may share this role
+(and choose how to do this). The
```
eteq, line 91:
```suggestion
(and it is at their discretion to decide how to divide up work). The
```

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,53 @@ posted on the pull request and also in a reply to your email.
...
+editor will start off by finding someone to carry out the main review of the
+package (hereafter the *reviewer*). In this sense the role of the editor is
+similar to that of a journal editor (and hence the role name). However this is
+only partially true, because here the editor may add their views to the review,
```
eteq, line 95:
```suggestion
similar to that of a journal editor (and hence the role name). However this analogy
is only partially accurate, because here the editor may add their views to the review,
```

affiliated/affiliated_package_review_process.md
```diff
@@ -82,41 +82,53 @@ posted on the pull request and also in a reply to your email.
...
+is not available in the project, reviewers can be requested from outside the
+project at the editor's discretion.
+
+The editor sends out an email to possible reviewers to find someone who is
+willing to review the package, using the template at the bottom of this
+document. Emails should only be sent to the next potential reviewer(s) - not
+many possible reviewers at the same time. The editor should let the potential
+reviewer know at this point that the editor may also be doing a review and
+synthesizing the results into a single review. Once a reviewer accepts, the
+editor send them an email with detailed review instructions (which includes a
```
eteq, line 111:
```suggestion
editor sends them an email with detailed review instructions (which includes a
```

eteq:
@astrofrog said he was fine with committing my suggestions and then merging, so here goes!

PR #91, opened 2020-05-11
Python 3 incompatibility is no longer acceptable

pllim:
It is May 2020 now.


eteq:
Good catch, @pllim, thanks!

PR #93, opened 2020-05-11
Add standard commit rights blurb

eteq:
This PR both adds a directory for messages like this and the content of a message the Coordination Committee generally sends as part of the process of giving someone write access to a coordinated repo.

Previously this was in a google drive shared with the coordinators, but I'm thinking if we are happy with it, it makes more sense to have it in an open visible place.


messages/README.md
```diff
@@ -0,0 +1 @@
+This directory holds emails or similar repeated or "frorm" emails that are part of some Astropy-specific process.  E.g., emails to people with new roles.
```
astrofrog, line 1:
```suggestion
This directory holds emails or similar repeated or "form" emails that are part of some Astropy-specific process.  E.g., emails to people with new roles.
```

pllim:
"the deputies should yield to the lead maintainer to make the final call" -- I thought there is no more lead vs deputy?

Also, add a link to https://docs.astropy.org/en/latest/development/workflow/maintainer_workflow.html ?

eteq:
Good points both, @pllim ! I added a commit addressing both - let me know if you're happy with them.

messages/core_write_access.md
```diff
@@ -0,0 +1,14 @@
+Hello {NAME},
+
+{some information about why they are getting access now}
+
+You can see the formal description of the sub-package maintainers role at http://www.astropy.org/team.html.  But there are some other guidelines to keep in mind, which I'm describing here.  In general you should only merge PRs that are in the domain of your subpackage(s).  This is a guideline rather than a firm rule: if you see something that's straightforward and obvious (e.g. a minor typo fix that's already been approved and the tests have passed but no one hit the merge button), you should feel free to help with that sort of thing even in other packages. Similarly, in areas where there's conflict/uncertainty, it's best to defer to a more senior maintainer (if there is one) or if absolutely necessary the Coordination Committee... but in cases where there's no obvious reason to question it (particularly if the lead is the PR author), it is also fine to merge things when it seems reasonable.
```
pllim, line 5: sub-package -> subpackage

pllim: e.g. -> e.g., (with comma)

messages/README.md
```diff
@@ -0,0 +1 @@
+This directory holds emails or similar repeated or "form" emails that are part of some Astropy-specific process.  E.g., emails to people with new roles.
```
hamogu, line 1: I suggest:
This directory holds emails or similar repeated "forms" that are part of some Astropy-specific process, e.g., emails to people with new roles.

eteq:
Thanks for the reviews, @astrofrog, @hamogu, and @pllim!  I think I've addressed your comments.

messages/core_write_access.md
```diff
@@ -0,0 +1,14 @@
...
+
+You can see the formal description of the subpackage maintainers role at http://www.astropy.org/team.html.  But there are some other guidelines to keep in mind, which I'm describing here.  In general you should only merge PRs that are in the domain of your subpackage(s).  This is a guideline rather than a firm rule: if you see something that's straightforward and obvious (e.g., a minor typo fix that's already been approved and the tests have passed but no one hit the merge button), you should feel free to help with that sort of thing even in other packages. Similarly, in areas where there's conflict/uncertainty, it's best to defer to a more senior maintainer (if there is one) or if absolutely necessary the Coordination Committee... but in cases where there's no obvious reason to question it (particularly if the lead is the PR author), it is also fine to merge things when it seems reasonable.
+
+Relatedly, if you haven't already, you should familiarize yourself with the maintainer guidelines: https://docs.astropy.org/en/latest/development/workflow/maintainer_workflow.html.  Please respond indicating you've seen these so we're sure you're aware of them.
```
bsipocz, line 7: the maintainer guidelines are actually extremely outdated and occasionally factually wrong. There is no way to push to master now, and some of the manual merges are also abandoned around ~v1.0. E.g. this section should be deleted altogether, rather than encouraged to be followed.
https://docs.astropy.org/en/latest/development/workflow/maintainer_workflow.html#integrating-changes-manually

we also don't have backport labels, etc.

bsipocz, line 7: Please do link to the basic developer guidelines, too, and to the git tutorials spelling out that rebase/squash and working with other people's fork as remote is commonplace tasks

messages/core_write_access.md
```diff
@@ -0,0 +1,14 @@
...
+
+You should also feel free to help with labeling new PRs or issues as they come in - that's a task all of the subpackage maintainers share, since it's only people with write privileges can do.  So don't feel at all shy about labeling these, even outside your subpackage(s).
```
bsipocz, line 9: maybe also add a feel free to "review PRs". We need more maintainers to review PRs without the fear to step on toes.

messages/core_write_access.md
eteq: Hmm.  I see your point, I hadn't actually read it in a while (perhaps ironically...).  Do you think it's better *not* to have maintainer guidelines then, and instead just link them to the basic workflow, or do you think having the maintainer guidelines is a good idea we just need to fix them?

(Either way you're right we should remove/replace with the basic dev workflow *for now* but the question is really about what to do in the astropy repo longer-term)

messages/core_write_access.md
```diff
@@ -0,0 +1,14 @@
+Hello {NAME},
+
+{some information about why they are getting access now}
+
+You can see the formal description of the subpackage maintainers role at http://www.astropy.org/team.html.  But there are some other guidelines to keep in mind, which I'm describing here.  In general you should only merge PRs that are in the domain of your subpackage(s).  This is a guideline rather than a firm rule: if you see something that's straightforward and obvious (e.g., a minor typo fix that's already been approved and the tests have passed but no one hit the merge button), you should feel free to help with that sort of thing even in other packages. Similarly, in areas where there's conflict/uncertainty, it's best to defer to a more senior maintainer (if there is one) or if absolutely necessary the Coordination Committee... but in cases where there's no obvious reason to question it (particularly if the lead is the PR author), it is also fine to merge things when it seems reasonable.
```
astrofrog, line 5: Instead of saying about the 'more senior maintainer' bit I wonder if it's better to say that it's important in difficult cases to get approval from *multiple* maintainers (this ties in to the policy @taldcroft has been writing). Actually maybe thinking about this this email should basically link to @taldcroft's policy document? (in other words, all guidelines should be in policies and you can link to them rather than repeat them?). Basically this email should probably be mostly links :)

taldcroft:
Looks fine to me subject to the changes already suggested.

messages/core_write_access.md
taldcroft: I wonder if some of the astropy core PR review in #100 and merge policy should be put directly into the maintainer guidelines.

hamogu:
I would suggest to break the lines (e.g. after 80 characters). That will make any future changes easier to see in the diff.

eteq:
In an out-of-band conversation, @taldcroft, @kelle, and I realized this should *also* include an expectation to reply and say the person agrees to the code of conduct and also acknowledge they are recognizing a representative of the community.

messages/core_write_access.md
eteq: While in a general sense I agree, I think it's important *in this email* to emphasize the value of judgement and community-building, since that's what the maintainers are there to do. So I totally agree with the "more senior maintainer" edit, but am less convinced we should lean on the policy doc links, since that takes agency away from the maintainers.  In any event, the policy docs don't exist yet, so my thinking is that I'll do the first half of this and we can defer the rest to when that #100 is in!

eteq:
Alright, I think I've updated based on all the comments and discussions. Only one thing I did *not* do (@hamogu's suggestion):

> I would suggest to break the lines (e.g. after 80 characters). That will make any future changes easier to see in the diff.

Normally I'd by happy to do that, but in this case it's usually a template for an email, and I find copying-and-pasting plain text into email clients is often more reliable without extra line breaks.  So I think it's better to leave it like this just because it's meant for email or similar messaging.

So this might be ready to merge (@taldcroft ?).

PR #155, opened 2020-10-16
Capture Finance Committee policy on issues

jdswinbank:


jdswinbank:
Unless any Finance Ctte members object (@eteq @kelle @hamogu), I'm going to merge this PR shortly on the basis that it's been hanging out here for a while now with no comments.

PR #170, opened 2020-12-22
Roadmaps

pllim:
A lot of work needs to be done here. I put in only a README for now to get the discussions going. 😬 

cc @astropy/coordinators @eblur


pllim:
@ceb8 is going to draft a new one based on Roadmap WG 2021 discussions.

PR #181, opened 2021-03-11
Funding 2021: Enhancements to gala and improved interoperability with other Astropy affiliated packages

adrn:
### Project Summary (copied from the proposal)

[Gala](https://gala.adrian.pw) is one of two specialized Galactic dynamics Astropy affiliated packages (with [Galpy](https://docs.galpy.org/)). Community consensus is that having two packages is beneficial to the community, as they naturally provide validation and checks on one another for implementations of detailed dynamical machinery. However, users who use both packages have requested to have enhanced interoperability between the packages. This proposal aims to fund a developer (the *developer*), mentored by the proposer (@adrn), to firstly improve interoperability between these two packages. Upon completion of this work, the *developer* would then work on improving interoperability between Gala and other dynamics community tools like AMUSE and Agama (not yet Astropy affiliated packages). Once this work is completed, the *developer* will have been introduced to much of the internals of gala: The third work package will leverage this expertise to work on new development within Gala.

### DIFF
--- a/finance/proposal-calls/2021-proposal/AdrianPriceWhelan.md
+++ b/finance/proposal-calls/2021-proposal/AdrianPriceWhelan.md
@@ -1,0 +1,106 @@
+### Title
+
+Enhancements to gala and improved interoperability with other Astropy affiliated
+packages
+
+
+### Project team
+
+Adrian Price-Whelan
+
+
+### Project Summary
+
+Gala is one of two specialized Galactic dynamics Astropy affiliated packages
+(with Galpy). Community consensus is that having two packages is beneficial to
+the community, as they naturally provide validation and checks on one another
+for implementations of detailed dynamical machinery. However, users who use both
+packages have requested to have enhanced interoperability between the packages.
+This proposal aims to fund a developer (the *developer*), mentored by the
+proposer (@adrn), to firstly improve interoperability between these two
+packages. Upon completion of this work, the *developer* would then work on
+improving interoperability between gala and other dynamics community tools like
+AMUSE and Agama (not yet Astropy affiliated packages). Once this work is
+completed, the *developer* will have been introduced to much of the internals of
+gala: The third work package will leverage this expertise to work on new
+development within Gala.
+
+
+### Project / work
+
+**Work package 1 (interoperability)**: Review and improve interoperability
+between Gala and Galpy
+
+The *developer* will first learn how to perform some core tasks in both
+[Gala](https://gala.adrian.pw) and [Galpy](https://docs.galpy.org/), such as
+numerically integrating an orbit, defining gravitational potential objects, and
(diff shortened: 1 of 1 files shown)


bsipocz:
From all the proposals, this feels to be an outlier to me. The proposed project is not related to any core astropy project activities (neither core nor coordinated libraries, associated infrastructure or teaching materials, etc.), but for a science, affiliated package. 
It also has the added risk of not having an identified personel, but proposing an open hire for a partial person. Another risk is that, as far as I see, the current situation of having multiple affiliated packages with very similar science goals is due to some disagreement within the science community. I very strongly feel that the astropy project should be neutral in these cases and shouldn't support one or the other, especially with funding.

So to summarize, I would prefer it if the project has clear boundaries. E.g. it takes responsibility for the coordinated packages, but don't treat affiliated packages preferably depending on whether their maintainer is someone on the astropy team or not.

adrn:
Good, yes, I think I agree with many of your points! Something I said on a call but didn't clearly state here: This was meant to be somewhat of an experimental proposal, to get feedback on the idea of funding this kind of work or to have the finance committee / community think about "out of the box" ideas (I agree that this is an outlier project!). I am still very interested in funding this work, and I don't think there are many other relevant opportunities for me in this context, but perhaps the outcome of this proposal could be to have a discussion here about where else I should look for this kind of funding.

I want to push back, though, on the idea that Astropy funding this work is somehow endorsing this package over another package: For one, the core point of this proposal is to support improving interoperability between the multiple community packages that exist! Another point is that it is somewhat irrelevant here that I'm on the astropy team: If Jo had also applied for funding to do similar work on Galpy, I think that could have also been taken seriously! Perhaps more relevant is that this proposal could have broadened scope a little bit to say that work would be done in both gala and galpy to improve interoperability in both, so that not just gala is singled out - fair enough to that!

bsipocz:
> I am still very interested in funding this work, and I don't think there are many other relevant opportunities for me in this context, 

Yes, that's tough. I think the first 2 WP are suitable, if not excellent GSoC etc projects, especially if there is any involvement included from the side of the other packages. And in fact, those are the ones about interoperability. 
The 3rd point reads very much as improvements to gala only, and could be one aspect of a science proposal, and I could see it being included in a  bigger NASA or NSF grant on the topic.

If the proposal is to consolidate the 2 package into one, then it's a bit different, but even then I think it's way too focused on one particular science sub field, and therefore is not within the realm of the current grant.

adrn:
Cool, thanks for the thoughts!

jdswinbank:
Please react to this comment to vote on this proposal (:+1:, :-1:, or no reaction for +0).

eteq:
Thanks for submitting this proposal @adrn - we are sorry to report that this didn’t gather enough support from the community to be accepted with the available funds. This PR will be closed without merging, but you should feel free to try again in a future round!

PR #193, opened 2021-04-05
Funding 2021: System for integration testing within the astropy project

bsipocz:
This is a project proposal to work on the ecosystem integration testing, a piece of infrastructure that is currently missing.


pllim:
The file is missing `.md` extension, so GitHub isn't rendering it. FYI.

dhomeier:
I think this would be a very useful improvement to the infrastructure. The scope is not quite clear to me yet – the (few) affiliated and coordinated packages I am somewhat familiar with have quite different testing regimes. So is one part of the project to encourage common testing standards, e.g. having tests both against astropy-dev and stable release (and LTS, perhaps)? And would that initially encompass a group of "pioneering" packages, or all willing to participate?

bsipocz:
Yes, I've already advocated for those testing standards for the years, and hope that I have parts in getting at least some of the ongoing astropy-dev testing established as well as broadening the RC testing in the wider community.

This project would make it mandatory to test with astropy-dev, stable, and minimal (==LTS?, hopefully we can standardise that, too, though I think it's reasonable to leave some leverage for coordinated packages to support older than LTS if they want to) for coordinated packages and encourage affiliated and other packages (especially that no build is necessary now using the nightly wheels).

And more importantly would also work out a way to close the loop, report back to astropy if/when there are any new regressions downstream. Working out the details would be part of the project (e.g. I could envision a CI job that is triggered by labels to test the downstream `main` branch for a PR. E.g. a PR with `nddata` label would run downstream `ccdproc` tests to ensure changes in e.g. `CCDData` don't break anything downstream, or a more realistic scenario, e.g. changes in `coordinates` are not breaking `specutils`. )
I would be happy with a system that is extended to well maintained downstream libraries with a team, rather than a sole maintainer, e.g. sunpy, gammapy, ones with a very quick feedback cycle, but would not think such a system would be suitable to extend to all the affiliates. 
Basically have to minimize the noise for the maintainers for astropy core to see only real regressions they are responsible for rather than downstream issues.

eteq:
Please react to this comment to vote on this proposal (:+1:, :-1:, or no reaction for +0)

jdswinbank:
Congratulations @bsipocz !  This proposal has the support of the Astropy Community and has been awarded $14,400 to implement.  This PR will be merged, and you can contact finance@astropy.org or reach out on Slack to arrange financial details (see more details in the email to astropy-dev).

PR #175, opened 2021-03-05
Funding 2021: Hire specutils/specreduce developer

hamogu:
*The Moore budget earmarks 22500 $ specifically for spectroscopy development. I submit this proposal to ensure that there is at least one spectroscopy proposal,and promise to oversee the work to the best of my abilities, but I'm happy to recind it if more detailed proposals are submitted.*

Spectroscopy is a weak spot for astropy, both specutils, and even more specreduce, need love. This proposal is to hire a developer with experience in spectroscopy from the astropy community to drive development forward.

PR #197, opened 2021-04-22
Funding 2021: Proposal for T. E. Pickering to work on specreduce

tepickering:
Proposal to support Timothy E. Pickering as a `specreduce` developer to help get it truly off the ground and usable by the community. This follows-up https://github.com/astropy/astropy-project/pull/175 and puts my hat in the ring to carry out at least some the work that is needed.

### DIFF
--- a/finance/proposal-calls/2021-proposal/tep_specreduce.md
+++ b/finance/proposal-calls/2021-proposal/tep_specreduce.md
@@ -4,7 +4,7 @@ Timothy E. Pickering
 
 ### Project Summary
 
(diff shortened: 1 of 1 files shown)


hamogu:
As interim finance committee member, I feel obliged to point out that the deadline for "Funding 2021" proposals has passed and I will therefore close this PR. That does not prevent you from being the -as yet unnamed- person that will do the work described in #175, but we can't have this as it's own PR at this point.

For purpose of fleshing out #175, the text added here is a great outline. Would you mind pasting it as a comment in #175 instead?
(I would have included it in the #175 text file, but the deadline to modify those files has also passed.)

PR #213, opened 2021-09-17
Add overhead rate to Finance Ctte policy

jdswinbank:


finance/finance-ctte-policy.rst
```diff
@@ -126,3 +126,12 @@ Access to project resources
    (Agreed `2020-06-26`__.)
 
    __ https://docs.google.com/document/d/193ERJ6wMNPaNjGSWbS94KHGnojpZ_zwXxWTi9L2etSo/edit#bookmark=id.9k7wgfaxak2d
+
+Rates and charges
+=================
+
+#. When contracting with an external entity for a project member's time, the standard overhead rate applied by Astropy is 30%.
```
hamogu, line 133: I wonder if we should avoid the word "standard overhead rate" here since that's a loaded term with a specific definition in some NASA circles. Just setting it in `"` might work, or `should include 30% of funding for general astropy use (similar to an overhead)"`.

jdswinbank: Hmm, interesting point.

I figured “overhead rate” was pretty standard terminology, so it seemed like a good idea to stick to that. The use of “standard” was to imply that we could also decide on “exceptional” rates on a case by case basis.

However, we should definitely avoid anything that might be an issue for NASA. How about:

> When contracting with an external entity for a project member's time, an additional 30% will normally be added to the hourly rate to cover general project expenses

?

hamogu: Sounds good to me.

astrofrog: It might make sense to find a more general term than 'hourly rate' since you may also decide to pay someone per deliverable at some point.

jdswinbank: Good idea! How about:

> When contracting with an external entity for a project member's effort, an additional 30% will normally be added to the basic cost to cover general project expenses

?

kelle: I like this language.

PR #215, opened 2021-09-23
Add draft of CoCo operating policies

mwcraig:
The intent of this is to let the community know how the CoCo intends to operate, to clarify some of the communication channels to/from the CoCo and to provide an outline for future CoCos as to how we operated.

### DIFF
--- a/policies/coco-operating-policies.md
+++ b/policies/coco-operating-policies.md
@@ -1,0 +1,120 @@
+# Astropy Coordination Committee Operating Policies
+
+This living document outlines the guiding procedures of the Astropy Coordination
+Committee.  Note that this is not meant to be exhaustive nor duplicative - for
+example, the running notes doc documents how it is meant to be used itself.
+This document instead is meant to describe procedure, and where necessary,
+policy, for how the CoCo operates.
+
+Note also: [APE0](https://github.com/astropy/astropy-APEs/blob/main/APE0.rst)
+lays out the role, election, and philosophy for the CoCo, and that content is not
+duplicated here.
+
+## Meeting Details
+
+All meetings are held remotely (using Astropy’s Zoom) except when CoCo members happen to be co-located.
+
+*Frequency*
+
+Meeting frequency can vary depending on required workload with the assent of
+the current CoCo.  The minimum expected rate is monthly
+
+*Quorum*
+
+There is currently no formal quorum for meeting or decisions. CoCo1 assumed 2,
+and CoCo2 generally informally required 3 although 2 was sometimes sufficient
+for urgent or unlikely-to-be-controversial decisions. Coco3 uses a quorum of
+3, however, members not present can leave a note in the agenda to ask to hold
+an item until they are present. At least positive 3 votes are needed for any
+decision.
+
+*Moderation*
+
+There is no dedicated chair.  Rather, the role of chair is taken on a
+per-meeting basis by a moderator who both prepares the agenda and moderates
+the discussion. The moderator role rotates through all CoCo members on a
+specified cycle (with swaps as needed to accommodate
+vacations/conflicts/etc).
+
+*Agenda/Notes*
+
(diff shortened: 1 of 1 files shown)


policies/coco-operating-policies.md
```diff
@@ -0,0 +1,120 @@
...
+duplicated here.
+
+## Meeting Details
+
+All meetings are held remotely (using Astropy’s Zoom) except when CoCo members happen to be co-located.
+
+*Frequency*
+
+Meeting frequency can vary depending on required workload with the assent of
+the current CoCo.  The minimum expected rate is monthly
```
hamogu, line 20:
```suggestion
the current CoCo.  The minimum expected rate is monthly.
```
I wonder if we ever get there. The finance committee started weekly meetings in the expectation that we'd be able to scale back at some point. That was two years ago and we've made it through the agenda with no items left twice in weekly meetings.

policies/coco-operating-policies.md
```diff
@@ -0,0 +1,120 @@
...
+
+*Quorum*
+
+There is currently no formal quorum for meeting or decisions. CoCo1 assumed 2,
```
hamogu, line 24:
```suggestion
CoCo1 assumed 2,
```
This sentence contradicts the stuff below. Coco three uses "3" for a decisions.

policies/coco-operating-policies.md
```diff
@@ -0,0 +1,120 @@
...
+## Outward Communication Channels
+
++ The running notes should be open.
+
++ To make announcements to the Astropy community, the CoCo generally relies upon
+  relevant mailing lists and forums.  When sending such announcements from the CoCo,
+  a CoCo member should be careful to sign a message as:
+
+        For the Astropy Coordination Committee,
+        <Name>
```
hamogu, line 70: Personally, I perfer it reversed:
```
Name
(for the Astropy Coordination Comittee)
```
but no strong feelings either way.

policies/coco-operating-policies.md
```diff
@@ -0,0 +1,120 @@
...
+Used for private, informal conversations among CoCo members. A new channel is created
+each time CoCo membership changes. There is a strong expectation of privacy and
+messages should not be shared without express permission of the sender.
+
+*Other Project Emails (e.g. confidential@astropy.org)*
+
+As with the coordinators list, there is generally an expectation of privacy from those
+who send to these emails, although that varies with the email.
+`confidential@astropy.org` has particular status in that it goes direct to the
+Project Ombudsperson.
```
hamogu, line 95: In fact, I would take that out her. confidential@ does not go to the Coco, so it should not be in the Coco operating procedures. We don't want people to think that the Coco can (or even would!) snoop on communication to the ombdbusperson.

policies/coco-operating-policies.md
adrn: Are you suggesting a change? Useful context but I think it's fine to leave the minimum expected rate as monthly or biweekly. No strong feelings though..

policies/coco-operating-policies.md
mwcraig: I've removed this

policies/coco-operating-policies.md
hamogu: The change I suggest is adding "." at the end of the sentence. The test is just a comment.

policies/coco-operating-policies.md
mwcraig: @kelle @eteq any strong feelings here?

kelle: I agree with Moritz but I don’t think this exact implementation detail needs to be prescribed. It can just be an  example implementation

kelle:
Looks good to me!

policies/coco-operating-policies.md
hamogu: No need to bikeshed. I think following the spirit of this is more important than the exact words anyway.

Issue #161, opened 2020-10-23
NASA ROSES-2020 Proposal for Astropy


jdswinbank:
NASA has announced a call for proposals for [Support for Source Tools, Frameworks, and Libraries](https://nspires.nasaprs.com/external/solicitations/summary.do?solId=%7b958CF134-D655-E512-B5AD-84501D14A0C1%7d&path=&method=init) as part of the [ROSES-2020](https://nspires.nasaprs.com/external/solicitations/summary.do?solId=%7bBCEE336B-D550-CCBA-1C8C-7A866DB06F45%7d&path=&method=init) funding programme.

This opportunity seems ideally suited to Astropy! The Interim Finance Committee hopes we can come together as a project to agree on our priorities and submit a fantastic proposal. To start that process going, we'd like to invite the community to use this issue to do some brainstorming — what are your suggestions and priorities for how we should respond to this call?

The deadline for submitting a “notice of intent” that we intend to submit a proposal is 19 November (the proposal itself won't be due until next year). Well before then, we'll come up with a more structured way of discussing these ideas and figuring out how we want to move forwards. For now, please just regard this as a free-for-all brainstorming session... and suggest something awesome!

eblur:
Yes! We should definitely do this, as either several separate proposals or together as one. Here are some areas that we could use some funding:

(1) Learn Astropy – hiring a dedicated technical writer to develop guides (along the lines of [Matt's CCD Reduction and Photometry guide](https://github.com/astropy/ccd-reduction-and-photometry-guide))
(2) Learn Astropy – hiring people to develop workshop curricula using research based practices
(3) Specutils – I would like to hire some one to develop tools for X-ray spectroscopy that are compatible with specutils. I have [already started this project](https://github.com/eblur/pyxsis) but no longer have the time to develop it for myself, beyond what I need for my own research. This is something I could consider submitting as PI of a separate proposal.

hamogu:
The call focuses on maintenance and that's where I see a deficit in the the core package. It's not just about the infrastructure, but in particular the older and most fundamental astronomy packages: fits and wcs. There are plenty of annoying bugs and reasonable feature requests for either, many of them have been open for years (at time of writing there are 150 open issues for fits and 60 for wcs). It's not sexy, but I think fixing those is absolutely necessary and won't ever be done without dedicated funding. Could possibly be done under the Moore grant, but if not, should definitely be done here.

pllim:
[...]

PR #218, opened 2021-10-22
Add proposals

eteq:
This adds a space for and info about the existing large grants to the Astropy project and their associated proposals..


kelle (approved):

kelle:
Since we're working on updating the docs, might be a good time to finish this one off as well.

jdswinbank:
Great! But a few comments:

- It'd be nice to make the filenames a bit more specific. At the moment, we only have one NASA ROSES proposal, but maybe that'll change in future so it probably makes sense to call the file something like `2020_nasa_roses` or similar.
- Documents should link to the GitHub issues on which the work to acquire the funds was carried out. So that's #161 for ROSES, etc.
- Do we really want this to be “successful proposals”? Don't we also want to keep track of proposals that were _unsuccessful_, so that we can learn from them? What about grants that weren't the result of a proposal process?
- Can we define “large”?

aaryapatil (approved):
Agree with @jdswinbank's comments above. Otherwise the text looks good - nice to have all the proposal documents in one place.

hamogu:
Suggestion: In addition to having our proposals available for astropy, put them here: https://www.ogrants.org/

Having them in out github repro or as a link to a google drive (as in https://github.com/astropy/astropy-project/issues/161 for the NASA proposal) *might* be enough for astropy community members to find them, but contributing them to a more general "open collected proposals" repository for other institutions or projects, who might have less experience with grant writing than our community, to find them. If the PI or finance committee agree, I can do the step of uploading them, but I don't want to do that without agreement, since it's not clear who have the copyright to the proposal. Probably all of us, but certainly the PI, so with their approval, I don't have to waste time on meditating on copyright law (not my preferred past time....)

jdswinbank (approved):
Sorry for taking so long to notice this hanging out at the bottom of my inbox.

jdswinbank:
Also added Moritz's suggestion to our upcoming meeting agenda.

pllim:
Does this need more follow up. I see approvals.

PR #230, opened 2021-12-23
Deprecate finance wiki

jdswinbank:
Addresses #224.

PR #231, opened 2022-01-03
Add step to procedure for new finance committee member

hamogu:
what it says in the title... one more step to write down so we don't forget (and don't forget when we remove members!)


aaryapatil:
Good catch! Based on all the steps taken till now to add me, this looks good.

aaryapatil (approved):

aaryapatil:
Quick question: The document title is `Interim Finance Committee Procedures` and the body of the text uses `interim` everywhere. I read #230 that the interim status was removed in September 2021. If I am understanding this correct, shouldn't we change `interim` here? Let me know if I have misinterpreted something 🙂

jdswinbank (approved):
LGTM. Thanks!

kelle (approved):

PR #233, opened 2022-01-20
Add step "add to mailing list" to policy for new roles

hamogu:


kelle (approved):

hamogu:
We talked about this in the Coco meeting and agreed on it already, so one approved review should be sufficient.

PR #234, opened 2022-01-27
Update process for adding role members

hamogu:
This adds an explicit step to check with the people currently
holding a role before inviting someone new.
In practice, we already that that (and nominations usually
come from current role members), but it's good
to write down so it's not forgotton.


mwcraig (approved):

kelle (approved):

mwcraig:
Merging, since we have two approvals now....

PR #235, opened 2022-01-28
Fix typo in draft massage to new maintainers

hamogu:


mwcraig (approved):
Small change and it looks fine, so merging...

kelle:
Not a big deal, but I think joining astropy-dev and joining the dev telecons should be two different sentences. 

Also, this would be a good place to point folks to the shared project calendar (which is admiringly not well advertised). I can open a new PR unless someone else wants to re-open and modify this one.

mwcraig:
Good ideas -- I think a new PR would be easier.

PR #269, opened 2022-07-08
Phase 3 Funding | Starkman & Foreman-Mackey | Units

nstarman:
For your consideration.

Signed-off-by: nstarman <nstarkman@protonmail.com>

### DIFF
--- a/finance/proposal-calls/cycle3/Starkman_&_Foreman-Mackey-Units.md
+++ b/finance/proposal-calls/cycle3/Starkman_&_Foreman-Mackey-Units.md
@@ -1,0 +1,99 @@
+### Title
+
+### Project Team
+
+Nathaniel Starkman (NS) and Dan Foreman-Mackey (DFM)
+
+
+### Project Description
+
+[`Quantity`][Quantity] is a backbone of Astropy, leveraging the power of [NumPy]
+[NumPy] and adding units. However, the computing landscape is changing. Many
+numpy-like libraries exist for different use cases: [Dask][Dask] for
+distributed computing, [CuPy][CuPy] for GPU, [xarray][xarray] for labeled
+data, [JAX][JAX] for autograd and GPU acceleration, [zarr][zarr] for compressed
+arrays, etc. Though these libraries offer a numpy-like API and interoperability
+protocols, `Quantity` is incompatible with them, which prevents users from
+choosing the most suitable library while also having units. In other words,
+users working on larger-than memory data or doing machine learning, among other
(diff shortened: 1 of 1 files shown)


eteq:
I for one love this project!  One question for @nstarman and/or @dfm : I note dask has not come up here other than as one of many examples.  I just submitted the much-less-fleshed out #283 (as an open call) - do you think that's actively in *conflict* with that? If so I would be fine deferring #283 to the future so you have some time to try this.

nstarman:
@eteq, glad to hear it!

We are not planning to directly implement Dask support in this proposed project, but the framework we will describe in the report is compatible with Dask and many other duck arrays. When the report is submitted and if the community likes it and an APE is made and implemented then Quantity + Dask, Sparse COO, xarray, pandas are all things we can implement -- Some, like Dask, more easily than pandas. I'm not sure what the funding timeline looks like, but perhaps #283 could be to fund someone to finish the reference implementation from this proposal, focusing on Dask. We believe we've solved how Quantity can work with any duck array and still have the current API (including `view`).

eteq:
Because the estimated total of all the proposals is a bit higher than the amount available for allocation, we suggest that you (@nstarman) provide a budget range - i.e., a "minimum for this project to be viable" and "maximum you could use effectively for this project".

kelle:

pllim:
How does this overlap with the dicussions at https://github.com/astropy/astropy/issues/13460 and https://discuss.scientific-python.org/t/unitsxd-units-across-domains-coordination/349 ?

nstarman:
Apologies, I miss this notification. I think this is very relevant. https://discuss.scientific-python.org/t/unitsxd-units-across-domains-coordination/349 links two related, but separate discussions:  how units work (multiplication, dimensions, etc), and how the array are stored. Both are necessary, especially if we want a unified approach. This proposal is to focus on the latter question and map out how Astropy can both support the growing community of array libraries and also remain backward compatible during this transition.

jdswinbank:
Congratulations!  The community has expressed support for this FR, and the Finance Committee has determined it can be funded.  This FR has been approved for $8,400, to be funded from the NASA grant.

@eteq is the Finance Committee member who will be supporting this project.  They will update this PR with the amount and then merge it, as well as being in contact with you to set up any necessary agreements and relevant Open Collective financial arrangements. They will also open a new issue where progress on this project can be discussed.

nstarman:
Thanks @jdswinbank for this update. I'm excited to receive this funding and grateful for Astropy's investment in this project.

PR #275, opened 2022-07-08
Phase 3 Funding: Improving Compressed Image performance and Dask integration in `io.fits`.

Cadair:
In this PR we propose development effort to improve the state of compressed image support in `astropy.io.fits` and to implement initial support for reading FITS files into Dask arrays.

This proposal is in two parts:
* **Part one**: Implement a tiled decompression/compression package to replace astropy's current use of `cfitsio`, to simplify the codebase and to open the door to substantial performance improvements.
* **Part two**: Write an initial implementation of loading FITS ImageHDUs directly in to `Dask` arrays (both compressed and uncompressed images. [Note that writing from a Dask array is already implemented.]

[Rendered Proposal](https://github.com/Cadair/astropy-project/blob/aperio_fits/finance/proposal-calls/cycle3/aperio_fits_dask.md)

### DIFF
--- a/finance/proposal-calls/cycle3/aperio_fits_dask.md
+++ b/finance/proposal-calls/cycle3/aperio_fits_dask.md
@@ -69,10 +69,34 @@ Some relevant issues:
 
 While this is one of the largest performance issues in `io.fits`, this project
 is also proposing to increase the integration of Dask with `io.fits` which will
-enable significant performance improvements when using FITS files with
-distributed compute.
+enable significant performance improvements when using FITS files with dask.
+
+This proposal is focusing on images in FITS files, so both uncompressed images
+and compressed images (which are stored in binary tables underneath), and
+proposes to add an option to `io.fits` to read both these types of FITS arrays
+directly into Dask arrays.
+
+While the proposal team has significant experience with reading various data
+formats into Dask arrays, for example FITS images and CASA images and tables.
+A proportion of the development time for this section of the proposal will be
+devoted to researching the most effective method of loading FITS files into Dask
+arrays.
+
+Currently it is
+[possible](https://github.com/sunpy/sunpy/issues/2715#issuecomment-413286821) to
+load an image into a Dask array, via the "delayed" functionality in Dask. In
+this case, the file is opened when reading a chunk of data from the array, and
+then closed again afterwards.
+This approach works well for a lot of use cases, but is complex, it would be a
+lot better if this were integrated into `io.fits` directly.
+
+For compressed images, Dask integration would allow you to process the
+compressed chunks of the image in parallel (either on a single machine or
+distributed), as if each compressed tile of the image was a dask chunk then it
+can be parallelised over using the various dask schedulers.
+
+This proposal is to get an initial implementation of both of these read use
+cases integrated into `io.fits`, and in the process to document any future
+improvements that could be made to enhance performance.
 
 ### Approximate Budget
-
-Money in return for the sacrifice of sanity required to rework large chunks of
-`io.fits`. /s
--- a/finance/proposal-calls/cycle3/aperio_fits_dask.md
+++ b/finance/proposal-calls/cycle3/aperio_fits_dask.md
@@ -1,0 +1,40 @@
+### Title
+Improve FITS compressed image support and Dask integration in `io.fits`.
+
+### Project Team
+Stuart Mumford
+Thomas Robitaille
+Drew Leonard
+
+### Project Description
+
+This project aims to increase performance of `io.fits` especially around large
+compressed image data.
+
+Currently when using tiled FITS compression the Astropy implementation loads
+all the tiles into RAM and decompresses them all.
+This is highly CPU and memory inefficient, and is one of the reasons Astropy is
+significantly slower at loading these types of files than the `cfitsio` package
+(which uses the same C library as Astropy for this currently).
+
+While this is one of the largest performance issues in `io.fits`, this project
+is also proposing to increase the integration of Dask with `io.fits` which will
+enable significant performance improvements when using FITS files with
+distributed compute.
+
+It's likely that any significant refactor of `CompImageHDU` will lead to an
+implementation not using the `fitsio` C library see
+[#3895](https://github.com/astropy/astropy/issues/3895) which also has the side
+effect of significantly reducing the compile time complexity of Astropy, as
+this is the only part of the code which uses `fitsio`.
+
+Some relevant issues:
+
+* https://github.com/astropy/astropy/issues/3895
+* https://github.com/astropy/astropy/issues/11633
+* https://github.com/astropy/astropy/issues/9238
+
+### Approximate Budget
+
+Money in return for the sacrifice of sanity required to rework large chunks of
+`io.fits`. /s
--- a/finance/proposal-calls/cycle3/aperio_fits_dask.md
+++ b/finance/proposal-calls/cycle3/aperio_fits_dask.md
@@ -97,9 +97,9 @@ package.
 
 Some relevant issues:
 
-* https://github.com/astropy/astropy/issues/3895
-* https://github.com/astropy/astropy/issues/11633
-* https://github.com/astropy/astropy/issues/9238
+* Compressed HDU enhancements [#3895](https://github.com/astropy/astropy/issues/3895)
+* Writing compressed FITS files is slow when multi-threaded [#11633](https://github.com/astropy/astropy/issues/11633)
+* Refactor FITS compressed image headers (CompImageHeader) [#9238](https://github.com/astropy/astropy/issues/9238)
 
 
 #### Dask Integration with reading FITS files
--- a/finance/proposal-calls/cycle3/aperio_fits_dask.md
+++ b/finance/proposal-calls/cycle3/aperio_fits_dask.md
@@ -131,7 +131,13 @@ lot better if this were integrated into `io.fits` directly. For example, somethi
 
 ```python
 hdulist = fits.open(filepath, use_dask=True)
-hdulist[0].data
+hdulist
+```
+```
+[<astropy.io.fits.hdu.image.DaskPrimaryHDU object at 0x7f1824914c10>, <astropy.io.fits.hdu.table.DaskCompImageHDU object at 0x7f1824914ee0>]
+```
+```python
+hdulist[1].data
 ```
 ```
 dask.array<reshape, shape=(4, 490, 1000, 2560), dtype=float64, chunksize=(1, 1, 1000, 2560), chunktype=numpy.ndarray>
@@ -148,3 +154,22 @@ cases integrated into `io.fits`, and in the process to document any future
 improvements that could be made to enhance performance.
 
 ### Approximate Budget
+
+For part one, we would request a total of 150h @ $150/hour broken up into:
+
+* 100h for removing the cfitsio dependency on `CompImageHDU`. This includes
+  reimplementing or adapting the C implementation of the tile (de)compression
+  algorithms, and modifying `CompImageHDU` to use the new astropy versions of
+  these algorithms rather than parsing all the data to `cfitsio` for loading.
+* 50h for implementing a new lazy-loading property on `CompImageHDU` which
+  allows users to load parts of the whole compressed image array without having
+  to load and decompress the whole array.
+
+For part two, we request another 100h @ $150/hour to implement prototype loaders
+for FITS Image HDUs making use of Dask. At the end of part two we expect to have
+at least one approach to efficiently loading FITS data into Dask, suitable for
+large scale parallelisation, and a plan of how these prototypes could be
+improved upon and merged into astropy core.
+
+Our minimal budget is the 150h requested for part one, which would mean leaving
+all the Dask work to another proposal.
--- a/finance/proposal-calls/cycle3/aperio_fits_dask.md
+++ b/finance/proposal-calls/cycle3/aperio_fits_dask.md
@@ -1,5 +1,5 @@
 ### Title
-Improve FITS compressed image support and Dask integration in `io.fits`.
+Improve FITS compressed image performance and Dask integration in `io.fits`.
 
 ### Project Team
 Stuart Mumford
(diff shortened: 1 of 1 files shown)


jdswinbank:
Because the estimated total of all the proposals is a bit higher than the amount available for allocation, we suggest that you (@Cadair ) provide a budget range - i.e., a "minimum for this project to be viable" and "maximum you could use effectively for this project".

Cadair:
@martindurant This is still a very early draft, but do you have any thoughts or recommendations on how you would go about adding dask support to reading FITS files in Astropy?

martindurant:
Yes, I will certainly have thoughts!

In the meantime, you might be interested in [this codec](https://github.com/fsspec/kerchunk/blob/main/kerchunk/codecs.py#L126) which can read variable-length table data from FITS, i.e., the heap, which would be a necessary prerequisite to loading the same data in parallel via zarr rather than astropy.io.fits. (The "rice" compressor isn't implemented in numcodecs, though, or - as far as I can tell - anywhere at all)

martindurant:
Some thoughts: 
- it is very reasonable to want to de decompression in parallel with dask, if you can release the GIL and overheads are reasonable. When we are comparing runtimes of ~100ms, overheads *will* be important, but up to 4-8x speedup might still be possible with the threaded scheduler
- more important is lazy/out-of-core or selective fetching of the data, both to cutdown on transfer bandwidth (I am thinking of remote loading) as well as improving CPU load
- probably remote loading can *only* work with a python implementation, being able to pass a file-like rather than a real local file handle, so I am fully behind this
- the structure of a single compressed image is perfectly aligned with a kerchunk representation, loading piecewise via a single decompression codec. So writing a references scanner for kerchunk would be be pretty simple if we can handle the compression codecs, and you would get remote/parallel/out-of-core fo free. This turns the problem on its head by using zarr as the IO layer instead of fits, but I would suggest astropy should have zarr as a supported data format backend. It would also be possible, but more work, to bundle a smart IO layer under fits.HDU (see, e.g., what fsspec.parquet does to preload the bytes ranges we know we will need, and then pass this prefilled file instance to the loader).

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+
+Currently when using tiled FITS compression the Astropy implementation loads
+all the tiles into RAM and decompresses them all.
+This is highly CPU and memory inefficient, and is one of the reasons Astropy is
+significantly slower at loading these types of files than the `cfitsio` package
+(which uses the same C library as Astropy for this currently).
+
+For example we can compare the loading times of `io.fits` vs `cfitsio`:
+
+Loading a whole array with astropy:
```
astrofrog, line 24: How big is the file? (both Gb but also array shape)

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+
+```
+In [44]: %timeit fits.getdata("VBI_L1_00656282_2018_05_11T14_25_05_466665_I.fits", hdu=1)
+183 ms ± 248 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
+```
+
+with cfitsio:
+```
+In [45]: %timeit hdu[1][:, :]
```
astrofrog, line 33: This should probably show all the code required to get here? (it's interesting if it's more verbose/tricky)

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+131 ms ± 182 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
+```
+
+loading one or more individual tiles with `cfitsio`:
```
astrofrog, line 37:
```suggestion
Loading one or more individual tiles with `cfitsio`:
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+22.3 µs ± 78.2 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
+
+In [50]: %timeit hdu[1][:10, :]
+235 µs ± 718 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
+```
+
+note that doing the same operations with astropy currently loads the whole file
+so takes the same amount of time as the first `getdata` call. This means that
+loading a single tile with astropy is approximately 8300x slower than with
+`cfitsio`!!
```
astrofrog, line 52: Maybe one ! is enough 😉

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+
+Issue [#3895](https://github.com/astropy/astropy/issues/3895) has been open
+since 2015 and describes a set of deep improvements for the compressed image
+support in `io.fits`. This includes implementing the compression natively in
+Python rather than using the `fitsio` C library to do it. This will have the
```
astrofrog, line 57: Should this say cfitsio when referring to the library?

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+side effect of significantly reducing the compile time complexity of Astropy, as
+this is the only part of the code which uses `fitsio`.
```
astrofrog, line 59:
```suggestion
the bundled cfitsio library could then be removed from the core package (as it was only used for the compression)
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+
+Some relevant issues:
+
+* https://github.com/astropy/astropy/issues/3895
```
astrofrog, line 63: Maybe include issue titles and make them links to the issues

finance/proposal-calls/cycle3/aperio_fits_dask.md
astrofrog: Also maybe say 'in this case' - bigger files will have an even bigger speedup

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+* https://github.com/astropy/astropy/issues/11633
+* https://github.com/astropy/astropy/issues/9238
+
+
+#### Dask Integration with reading FITS files
+
+While this is one of the largest performance issues in `io.fits`, this project
```
astrofrog, line 70:
```suggestion
While handling compression is one of the largest performance issues in `io.fits`, this project
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+is also proposing to increase the integration of Dask with `io.fits` which will
```
astrofrog, line 71:
```suggestion
is also proposing to increase the integration of Dask with `io.fits` in general which will
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+enable significant performance improvements when using FITS files with dask.
```
astrofrog, line 72:
```suggestion
enable significant performance improvements across the board when using FITS files with dask.
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+
+This proposal is focusing on images in FITS files, so both uncompressed images
+and compressed images (which are stored in binary tables underneath), and
+proposes to add an option to `io.fits` to read both these types of FITS arrays
+directly into Dask arrays.
+
+While the proposal team has significant experience with reading various data
+formats into Dask arrays, for example FITS images and CASA images and tables.
```
astrofrog, line 80: Incomplete sentence? Remove 'While'?

Maybe mention that we made it possible to write dask arrays to file efficiently, and link to the PR?

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+A proportion of the development time for this section of the proposal will be
+devoted to researching the most effective method of loading FITS files into Dask
+arrays.
+
+Currently it is
+[possible](https://github.com/sunpy/sunpy/issues/2715#issuecomment-413286821) to
+load an image into a Dask array, via the "delayed" functionality in Dask. In
```
astrofrog, line 87:
```suggestion
load an image from a FITS file into a Dask array, via the "delayed" functionality in Dask. In
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+directly into Dask arrays.
+
+While the proposal team has significant experience with reading various data
+formats into Dask arrays, for example FITS images and CASA images and tables.
+A proportion of the development time for this section of the proposal will be
+devoted to researching the most effective method of loading FITS files into Dask
+arrays.
+
+Currently it is
+[possible](https://github.com/sunpy/sunpy/issues/2715#issuecomment-413286821) to
```
astrofrog, line 86:
```suggestion
[possible](https://github.com/sunpy/sunpy/issues/2715#issuecomment-413286821) but not trivial to
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+load an image into a Dask array, via the "delayed" functionality in Dask. In
+this case, the file is opened when reading a chunk of data from the array, and
+then closed again afterwards.
+This approach works well for a lot of use cases, but is complex, it would be a
+lot better if this were integrated into `io.fits` directly.
+
+For compressed images, Dask integration would allow you to process the
```
astrofrog, line 93:
```suggestion
For compressed images, Dask integration would allow users to process the
```

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+In [50]: %timeit hdu[1][:10, :]
+235 µs ± 718 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
+```
+
+note that doing the same operations with astropy currently loads the whole file
+so takes the same amount of time as the first `getdata` call. This means that
+loading a single tile with astropy is approximately 8300x slower than with
+`cfitsio`!!
+
+Issue [#3895](https://github.com/astropy/astropy/issues/3895) has been open
```
astrofrog, line 54: This paragraph is good but I think also missing that it should say explicitly that aside from writing the compression in Python one of the big things is obviously to use dask to do on-demand decompression of chunks (as it is right now, this sub-section is missing a 'and we will therefore do this with dask' - I think it would also be good to somewhere mention in a couple of sentences what is dask and why it can help.

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+* https://github.com/astropy/astropy/issues/9238
+
+
+#### Dask Integration with reading FITS files
+
+While this is one of the largest performance issues in `io.fits`, this project
+is also proposing to increase the integration of Dask with `io.fits` which will
+enable significant performance improvements when using FITS files with dask.
+
+This proposal is focusing on images in FITS files, so both uncompressed images
```
astrofrog, line 74: You mention images but what about cubes? (maybe re-word to make it clear that would be covered)

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+this is the only part of the code which uses `fitsio`.
+
+Some relevant issues:
+
+* https://github.com/astropy/astropy/issues/3895
+* https://github.com/astropy/astropy/issues/11633
+* https://github.com/astropy/astropy/issues/9238
+
+
+#### Dask Integration with reading FITS files
```
astrofrog, line 68: I'm not sure about the way this is divided into 2 sub-sections currently because the second one somewhat also includes the compression stuff, so it's a bit higher level perhaps?

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+
+This project aims to increase performance of `io.fits` especially around large
+compressed image data.
+
+#### Compressed Image Performance
+
+Currently when using tiled FITS compression the Astropy implementation loads
+all the tiles into RAM and decompresses them all.
+This is highly CPU and memory inefficient, and is one of the reasons Astropy is
+significantly slower at loading these types of files than the `cfitsio` package
```
astrofrog, line 19: link to cfitsio? Also make it clear if you mean cfitsio (the C library) or fitsio (the Python library). I think here maybe you actually mean fitsio?

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
+### Title
+Improve FITS compressed image support and Dask integration in `io.fits`.
```
astrofrog, line 2: 'performance' instead of 'support'?

finance/proposal-calls/cycle3/aperio_fits_dask.md
astrofrog, line 54: I wonder if you should summarize some of the main points here and be more explicit about what you are proposing *we* do here

finance/proposal-calls/cycle3/aperio_fits_dask.md
```diff
@@ -0,0 +1,102 @@
...
+devoted to researching the most effective method of loading FITS files into Dask
+arrays.
+
+Currently it is
+[possible](https://github.com/sunpy/sunpy/issues/2715#issuecomment-413286821) to
+load an image into a Dask array, via the "delayed" functionality in Dask. In
+this case, the file is opened when reading a chunk of data from the array, and
+then closed again afterwards.
+This approach works well for a lot of use cases, but is complex, it would be a
+lot better if this were integrated into `io.fits` directly.
```
astrofrog, line 91: Maybe could give some kind of pseudo-code example such as:

```
>>> hdulist = fits.open('bigfile.fits', use_dask=True)
>>> hdulist[0].data
<dask thingy>
```

or something like that, just to explain what 'integration' means.

Cadair:
@martindurant Just hypothetically, what would astropy need to expose for kerchunk to be able to "just" call astropy to expose a zarr array? @astrofrog and I were discussing different approaches to this problem today, and I think we were discussing implementing something fundamentally similar to how kerchunk has implemented FITS support but providing an API more familiar to current `astropy.io.fits` users. Obviously, if we were to end up taking that path, it would be redundant for both of us to carry around similar implementations, so knowing how to combine them would be great.

martindurant:
I think having astropy be able to read from zarr should be very simple, and kerchunk/dask would come for free. Whether it would want to export the same hdu/header interface as fits is another matter. Should check up on what asdf does right now.

pllim:
Did someone say asdf ? cc @WilliamJamieson

kelle:

jdswinbank:
Congratulations!  The community has expressed support for this FR, and the Finance Committee has determined it can be funded.  This FR has been approved for $22,500, to be funded from the NASA grant.

@aaryapatil  is the Finance Committee member who will be supporting this project.  They will update this PR with the amount and then merge it, as well as being in contact with you to set up any necessary agreements and relevant Open Collective financial arrangements. They will also open a new issue where progress on this project can be discussed.

bsipocz:
Why is this not given the full amount, while proposals with less support got the full request?

Cadair:
It is disappointing that the second half hasn't been funded (as that's the "flashy" part), but the first half is still required ground work, for both the second half of this proposal and for kerchunk to support tile compression etc.

eteq:
@Cadair - I think this was a mis-understanding of the proposa amountl!  That is, I think the intent was to fund this whole proposal, not just the first part, and it was mis-understood as the 150 hours being the requirement for *both* parts. Will confirm that to be sure but my recollection is that we intended to fully fund this proposal.

Cadair:
Ok @eteq that sounds great. (There's a non-zero chance we find ourselves with less time anyway :see_no_evil: )

astrofrog:
@aaryapatil - we have now completed the first bullet point in part 1 of the proposal which was to remove (as much as possible) CFITSIO from astropy:

* https://github.com/astropy/astropy/pull/1425
* https://github.com/astropy/astropy/pull/14252

These have both been merged (in addition to simplifying/removing a lot of FITS-related code, this works also speeds up compilation of the core package by 40% or so, so it has had benefits beyond just FITS).

We ended up using 138h for this work (invoiced today) instead of the original 100h we budgeted, so we now only have 12h left for the remainder of part 1. We are wondering if the finance committee would be happy to approve the additional 100h of funding we originally requested, as this would allow us to complete part 1 and also do a big fraction of part 2? (see https://github.com/astropy/astropy-project/pull/275#issuecomment-1281315500 for more details)

hamogu:
Are you sure you mean https://github.com/astropy/astropy/pull/1425 ? That was merged in 2013... Probably a typo of the PR number?

astrofrog:
Oops deleted a character by mistake when reformatting my comment 😅 - fixed now

aaryapatil:
This looks great. Thanks, @astrofrog! Could you please post your update on the issue (#316) instead of on the PR. We aim to use the issue for keeping track of the progress. We will also get back to you about the the additional 100h of funding (hopefully next week).

PR #281, opened 2022-07-08
Phase 3 Funding: Specreduce development

eteq:
Note @keflavich and @tepickering may want to sign onto this, or say they plan to do some of this, but I figured I'd start with an open call since I doubt they have capacity for all of this work!

### DIFF
--- a/finance/proposal-calls/cycle3/specreduce-FR.md
+++ b/finance/proposal-calls/cycle3/specreduce-FR.md
@@ -2,7 +2,7 @@
 Continue specreduce (and specutils) development
 
 ### Project Team
-This is an open hire, to be overseen by Erik Tollerud (@eteq), with likely contributions from Tim Pickering (@tepickering)
+This is part work by @tepickering and partly an open call to be overseen by Erik Tollerud (@eteq)
 
 ### Project Description
 This is to continue development that was begun in the previous cycle on building out pieces of specreduce.  Some specific areas that could be tackled by a developer in the coming cycle include:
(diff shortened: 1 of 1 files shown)


kelle:
Because the estimated total of all the proposals is a bit higher than the amount available for allocation, we suggest that you (@eteq ) provide a budget range - i.e., a "minimum for this project to be viable" and "maximum you could use effectively for this project".

eteq:
A question for finance committee to consider: this proposal is partly an open call and partly a specific developer. Should of be split between two proposals, one for each? I won't do that now given immunity sure how best to proceed, but the finance committee should discuss what the right pathway is here.

kelle:
Aarya and I discussed and we think it's fine for this to stay as one combined request.

aaryapatil:

mhvk:
Hard to judge this without some context of the overall landscape of spectroscopic reduction software -- the state here seems very elementary. Is this not at the start-of-astropy state, where really people maintaining different packages should get together and merge their efforts?

hamogu:
The problem is that that "getting-together" has not happened and (based on conversations we've had in the past), may not happen because existing packages serve specialized interests (e.g. work for this telescope, provide a fully automatic pipeline). One strategy is to build out specreduce enough that is can serve as a scaffolding for other packages. For example, I might want to do the trace-finding with DRAGONS and the optimal extraction with pyREDUCE. Having specreduce at least as basic, elementary scaffolding would allow the developers to make (at least initially) just single blocks available (e.g. trace-finding in DRAGONS)  without the need to have each and every step converted to some common API. It might also convince people as a proof-of-concept that this works.

Not ideal, and not ideal use of money, but that *might* work.

mhvk:
Yes, the worry is that nobody will use the scaffolding...

PR #284, opened 2022-07-08
Phase 3 Funding:  Fill vacant community engagement coordinator/community manager role

eteq:
An open call to hire someone into the community engagement coordinator role (see the text for a bit more justification on how this makes sense as an open call).

I included @kelle and @adrn based on earlier discussions with them, although they are welcome to bow-out if I misunderstood!

PR #286, opened 2022-07-08
Phase 3 Funding | Workshop Coordinator

kelle:
proposal to hire a dedicated workshop coordinator. Could be folded into Community Manager role (#284) if we wanted.

### DIFF
--- a/finance/proposal-calls/cycle3/workshop-coordinator.md
+++ b/finance/proposal-calls/cycle3/workshop-coordinator.md
@@ -1,0 +1,16 @@
+### Title
+Workshop Coordinator
+
+### Project Team
+This would be an open hire but would be supported by existing Learn Team.
+
+### Project Description
+A person to focus on workshops and in-person educational efforts. 
+- Coordinate Winter AAS workshop.
+- Expand workshops to more conferences.
+- Assess sucess of workshops, figure out metrics, design surveys.
+- Recruit more facilitators.
+
+
+### Approximate Budget
+Roughly 8,000 USD. Not sure.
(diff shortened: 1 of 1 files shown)


kelle:
Because the estimated total of all the proposals is a bit higher than the amount available for allocation, we suggest that you provide a budget range - i.e., a "minimum for this project to be viable" and "maximum you could use effectively for this project".

hamogu:
I think this proposal would benefit from a few sentences on what the future role of workshops should be. Are we still seeing an influx of new users that had no contact to Python before? Do workshops develop into more specialized packages (e.g. a ccdproc/photutils workshop)? Are we expecting to get contributors this way (and if so, what needs to be done to make that happen)? What about online-only workshops to improve access for non-US astronomers?

I understand that it might be part of this role to answer these questions, but a least a few initial thoughts would be useful. I have not attended an astropy workshop, so I have no idea which demographic they have served in the past.

kelle:
Great questions! Thanks for the feedback Moritz, will do.

aaryapatil:

astrofrog:
I personally think it would be better to combine this with the community engagement coordinator, as then a single person could spend more time and there are clearly going to be links about community engagement and workshops.

kelle:
Congratulations!  The community has expressed support for this FR, and the finance committee has determined it can be funded.  This FR has been approved for $12,000, to be funded from the Moore grant.

@kelle is the Finance Committee member who will be supporting this project.  They will update this PR with the amount and then merge it, as well as being in contact with you to set up any needed agreements and relevant Open Collective financial arrangements. They will also open a new issue where progress on this project can be discussed.

PR #291, opened 2022-07-26
Add proposed extension for Brasseur Astroquery maintenance

ceb8:
This is a request for an extension for the Cycle 2 Astroquery maintenance work carried out by Clara Brasseur through Aperio Software.


eteq:
The finance committee approved this extension - merging!

PR #298, opened 2022-09-12
added instruction for future funding calls

hamogu:
This PR is the result of me readinf the cycle 3 project description. In some cases, I really did not know how to vote because it was unclear to me who benefits from this proposal. For example, for the AstropyLearn proposal a little more specific info (e.g. page views, use cases, audiences) would have helped me to understand if this will be money well spend.

The same is true for a few other proposals in cycle 3; so I tried to add a sentence so that people who write future funding requests can keep that in mind. 

Since it's not appropriate to change the doc in the cycle3 directory at this stage, I've copied the file to the template directory where it will be found next year.

### DIFF
--- a/finance/proposal-calls/templates/project_template.md
+++ b/finance/proposal-calls/templates/project_template.md
@@ -1,0 +1,24 @@
+### Title
+
+### Project Team
+Describe who is expected to work on the project. If work will be undertaken by an open hire instead of (or in addition to) other project members, please indicate that here.
+
+### Project Description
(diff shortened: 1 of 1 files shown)


pllim:
Alternately, you could also implemented this as a GitHub issue template. See examples in https://github.com/astropy/astropy-project/tree/main/.github/ISSUE_TEMPLATE

pllim:
We have templates now, so I don't think we need this anymore.

https://github.com/astropy/astropy-project/tree/main/.github/ISSUE_TEMPLATE

PR #348, opened 2023-06-16
Funding request: ASDF-Astropy I and II

nden:
Request for funding for work on serializing astropy and astropy-coordinated objects to ASDF.

### DIFF
--- a/finance/proposal-calls/cycle3/asdf-astropy1.md
+++ b/finance/proposal-calls/cycle3/asdf-astropy1.md
@@ -1,0 +1,38 @@
+### Title
+ Expanding Astropy and Astropy-affiliated objects to be serialized to ASDF - Part I
+
+### Project Team
+- Elizabeth Madison: Approximately 1.5 years of scientific python programming experience; previous experience calibrating HST/WFPC camera and image analysis of resulting orbital data; much experience with image analysis and distortion calibration and projection onto planetary coordinate systems, particularly with regard to Jupiter and Saturn (CV upon request)
+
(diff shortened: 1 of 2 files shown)


eteq:
One clarifying question, @nden: I noticed the budget are the same for both parts, is that the budget for *both* parts together, or is the total request here $38400 ?

nden:
It's two different proposals with different budgets. I split the proposal roughly in two equal parts with the second depending on the success of the first.

jdswinbank:
@nden — Could you expand a little on the basis for the budget? Specifically, is the $19.2k per half estimated on the basis of a number of hours (at the end of which the work stops), or on delivering the finished product (regardless of how many hours of work that takes)? Thanks!

jdswinbank:
> @jdswinbank Does this answer your question?

Yes, that's great. Thanks!

kelle:
Just want to confirm that the individual completing the work would be funded as an Independent Contractor, correct?

pllim:
Hi. This request has been forward to CoCo from the Finance Committee. May CoCo request the CV of Elizabeth Madison? Thanks!

nden:
@kelle If "Independent Contractor" means not affiliated with an institution, then yes. She is an independent contractor.
@pllim We 'll forward the CV to CoCo.

jdswinbank:
Hi folks,

The size of the funding allocation being requested here is outside the remaining scope of the Moore project budget. This request would therefore have to be supported by funds from our NASA ROSES grant. On that basis, the CoCo have decided that this work should not be singled out for special approval at this stage. Instead, we'd like to encourage you to resubmit this for vote by the project members as part of our next regular call for funding. We don't have a timescale for that yet, but I informally expect it to be sometime in the next few months.

PR #392, opened 2024-03-01
Cycle 4 funding: astroquery (AAO-MQ)

NuriaLorente:
AAO submission for Cycle 4 funding round. Two proposals are submitted for consideration: 
- AAO-MQ-astroquery.md - Making Australian astronomical datasets accessible
- AAO-MQ-specutils.md - Realising the full potential of specutils via an extensible interface for community contributed algorithms.


kelle:
Also, these two proposals are quite different and I think would be more easily discussed and voted upon as two separate pull requests. Could you please open a 2nd one for one of the projects and adjust this one accordingly?

aragilar:
I've split off the specutils proposal into #399, but I can't change this one (I lack the permissions on github to do that).

aragilar:
@bsipocz @ceb8 @keflavich What's your opinion on where the cone of darkness algorithm should be placed within the astropy ecosystem? Currently our plan is to add the algorithm to astroquery directly, but we'd be open to adding the algorithm to astropy, and adding in a helper function/mixin class which handles the queries and wraps the algorithm to astroquery.

bsipocz:
There shouldn't any algorithms be added to astroquery, we only have modules that provide access to external services and data archives but there should be no value added functionality in the library itself (no algorithms, or any extra functions that provide stuff not available directly at the archives). Any such thing should live in another package that uses astroquery for data access.

ceb8:
Agree with @bsipocz, a good rule of thumb is that if you can envision a user importing astroquery to use a bit of functionality that does not interact with an external service, it doesn't belong in astroquery.

NuriaLorente:
> Could you clarify how the funding would work? Would the contributors be paid individually as independent contractors? Or do you envision the funding coming to AAO? If going through AAO, are there any overhead costs which need to be included?

Hi @kelle, we envision the funding coming to AAO, thus allowing members of the Research Data & Software team to become more involved in astropy development as part of their duties, rather than doing this "in their spare time". The AAO overhead costs are included in the quoted hourly rate, so the amount in the proposal is the total amount of funding requested.

eteq:

pllim:
I'd love to see more involvement from Down Under! 

As for funding, will there be other sources of income, e.g., from Australian observatories or IVOA?

> There shouldn't any algorithms be added to astroquery

What about `pyvo`?

bsipocz:
This particular one I think is also a no for PyVO as it has nothing to do with obtaining the data but doing something value added on top of the obtained data. So, the natural place for it would be 1) astropy core if it has a wide enough target audience (probably a no), or 2) an affiliated library, either an existing one or a new one that is topically appropriate.

bsipocz:
And as an astroquery maintainer, I cannot emphasize it enough how much I endorse this proposal as a way for bringing in a new archive with a new community. 
I suppose the answer above is that it's difficult to get funding, but I hope that with this seed funding for the initial implementation what actually will happen that there will be a long term commitment from the archive side to keep the module maintained.

bsipocz:
Not sure whether it's useful or not and where it should go,  my support is for the part of adding a module for the archive without the algorithm.

jdswinbank:
I'm writing on behalf of Astropy's Finance Committee regarding the outcome of this funding request. We are pleased to be able to let you know that, following consultation with the community, we are able to approve this request. We can currently fund the Year 1 amount of US$ 32,550.00 to carry out the project. Funding beyond this amount will be contingent upon the availability of funds. (We will be using the full budgets of all of the approved requests to craft future grant and funding proposals.) We assume you will be using this full year 1 budget, if that is not the case please contact us immediately.

Ana Gabela and I will be your contacts on the Finance Committee to facilitate this award. Please get in touch with us if you have any questions or concerns. Please do not reach out to NumFOCUS directly. Also, after merging this PR, I will open an issue where we can discuss progress on this project.

In addition, new to this funding cycle, is the assignment of a COTR (Contracting Officer's Technical Representative) to each funded project. This concept is borrowed from government funding agencies, although it is to be stressed that Astropy's goal is to make the COTR role as low-overhead as possible. The COTR’s primary responsibility is to make sure the work is happening at the expected pace and, if necessary, to be a liaison between the funded project and the Finance Committee or CoCo. The COTR for your project will be assigned shortly and we’ll also be sending out more details about how we see this working.

The next steps are:

* Please reach out to your institution’s grant office to identify the relevant contact for your subaward. Make sure they know that NumFOCUS is the entity they will be working with for the subaward. 
* When your institutional contact has communicated to you the local process, let me know and we will identify the specific next steps required on your institution’s end. We will also start the process with NumFOCUS.

Congratulations — we are really looking forward to seeing you put these funds to good use!

kelle:
The Finance Committee has approved both years of this request to be funded by the NASA ROSES24 grant. Amount: $65,100.

PR #394, opened 2024-03-01
Cycle 4: Cosmology

nstarman:


mhvk:
This seems a nice idea, but a possible worry is that it will be very hard to find a suitable person - who has to be good at cosmology as well as at coding - since otherwise supervision will take nearly as much time as doing the work.

nstarman:
Hi @mhvk, thanks for the comment. I've updated the proposal to better explain why the software engineer doesn't actually need a deep understanding of Cosmology, just a good ability to distill API from the structures of codes. This is really more of a time-to-review + structural analysis task that happens to be on a Cosmology subject.

kelle:
Consider adding a minimum budget.

eteq:

nstarman:
> Consider adding a minimum budget.

Thanks! Done.

pllim:
I feel like an APE is needed first before this can happen.

Also, how does this play with other "unified API" campaigns out there? Example:

* https://github.com/astropy/astropy/issues/16286
* https://data-apis.org/array-api/latest/

nstarman:
> I feel like an APE is needed first before this can happen.

For which part of the proposal?

> Also, how does this play with other "unified API" campaigns out there? Example:

It plays very nicely! When `Quantity` can support other array types, then it's a goal for Cosmology to support all the same types. This shouldn't be difficult and made easier with the proposed architecture.
 
> * [Use a neutral format to have lossless interface with JSON, Xarray, scipp, pandas astropy#16286](https://github.com/astropy/astropy/issues/16286)
> * https://data-apis.org/array-api/latest/

pllim:
> implement the The Cosmology API

Seems like this is a very significant change and should require APE?

nstarman:
> > implement the The Cosmology API
> 
> 
> 
> Seems like this is a very significant change and should require APE?

Not that large a change. Essentially 100% backwards compatible.
More a thoughtful re-org and cleanup of the components. Some functions gain an optional second argument.

AnaGabela:
Hi Nathaniel,

I'm writing on behalf of Astropy's Finance Committee regarding the outcome of your recent Funding Request.

We are sorry to report that your proposal could not be funded during this funding Cycle. We will be closing the FR as a result.

While the community vote was the primary driver of this, it is important to note that this likely does not reflect anything about you as a member of the Astropy Community. Rather, it reflects the specific balance of projects proposed this cycle, the available funds, and the priorities of this cycle’s funding sources relative to others.

We encourage you to read feedback in the thread above or other feedback you might have received via other means, and consider an FR for the next cycle with whatever modifications you think would help. You are also welcome to reach out to the Finance Committee for feedback if you would like.

Ana- on behalf of the Astropy Finance Committee

PR #395, opened 2024-03-01
Cycle 4 Funding: Spectroscopy Development Busy Week

kbwestfall:
Proposal requests funding to support a busy week for ~20 developers to push through development priorities for specutils and specreduce.

PR #399, opened 2024-03-14
AAO-MQ specutils Cycle 4 proposal

aragilar:
This is the split-out version of the specutils proposal in #392 for easier commenting.

On the questions raised in https://github.com/astropy/astropy-project/pull/392#issuecomment-1988617833, @NuriaLorente will provide details soon.

### DIFF
--- a/finance/proposal-calls/cycle4/AAO-MQ-specutils.md
+++ b/finance/proposal-calls/cycle4/AAO-MQ-specutils.md
@@ -35,12 +35,12 @@ can be involved if needed:
 The latest version of specutils includes useful common analysis tasks for
 astronomical spectra. 
 To realise its potential as a fully-fledged contributor to scientific workflows,
-spectutils would greatly benefit from the addition of a generic capability for
+specutils would greatly benefit from the addition of a generic capability for
 advanced users to supply their own tailored algorithms (i.e. code) to be used in place of
 the defaults. 
 We would like to add this capability to specutils, allowing users to apply
-advanced analysis and algorithms to specutils spectra via a regularlised
-interface.  This interface would be analagous to how astroquery provides a
+advanced analysis and algorithms to specutils spectra via a regularised
+interface.  This interface would be analogous to how astroquery provides a
 regularised interface to access VO and other web services.  It will provide the
 opportunity for the community to contribute their own algorithms, e.g. relating
 to large community spectroscopic surveys, or other generic methods. This would go
--- a/finance/proposal-calls/cycle4/AAO-MQ-specutils.md
+++ b/finance/proposal-calls/cycle4/AAO-MQ-specutils.md
@@ -69,13 +69,25 @@ We are keen to make substantial contributions to specutils in the future.
 
 ### Approximate Budget
 This funding application is to cover staff time. We estimate 630 hours of effort
-are required to complete this project, which at an hourly rate of USD$140/hour 
-gives a budget of USD$88,200.
+are required to complete this project, which at an hourly rate of USD$155/hour
+gives a budget of USD$97,650.
 
-This effort estimate will be refined, and the minimum useful time to carry out
-some of the proposed work will be determined and added here during the iteration
-period.
+#### Minimum useful time
 
+The absolute minimum useful time on this project would be 210 hours, which would
+only allow us to implement the FFT-based redshifting. However, no code
+templating capability would be included, which undercuts the value of the
+redshifting code as an important example of a case study algorithm. Furthermore,
+the redshifting code would be a minimal implementation with no integration with
+any existing specutils or astropy classes, no support for the conversion to
+logspace of templates or spectra as is required by the FFT algorithms. Nor would
+there be any time for astronomer-focused documentation, making it hard for
+astronomers to use the code. While this very stripped back version may provide
+some value, we would strongly suggest that funding the project in its entirety
+will give much better value for the astropy project, and for astronomer-users of
+specutils.
+
+At the aforementioned rate of USD$155/hour, this would be USD$32,550.
 
 ### Period of Performance
 We expect the period of performance to be two years.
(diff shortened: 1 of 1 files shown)


aragilar:
@nmearl @rosteen @keflavich @eteq Do you see any issues with the proposal/anything you'd be concerned about going into specutils?

rosteen:
> @nmearl @rosteen @keflavich @eteq Do you see any issues with the proposal/anything you'd be concerned about going into specutils?

Sorry for the delay replying, I was on vacation - I have one clarifying question. Is this intended primarily for users to replace algorithms on the fly / locally, or for users to add algorithms to some central repository (or both)? For the central repository case, I don't fully understand how this framework differs from/improves upon simply contributing algorithms through a Github pull request to specutils. I can definitely see how it would be useful at runtime to be able to say "do template comparison, but use this algorithm under the hood instead of the default specutils one."

aragilar:
It's envisioned that this will be to allow users to build up components locally, and hopefully these will get upstreamed (in a similar way to how loaders work in specutils currently). I can see the more experimental stuff may take longer to upstream (especially ensuring the code has sufficient tests to ensure that it does not regress), but the idea is to try to make it so that adding/switching algorithms does not imply restructuring/rewriting code. This would then be able to be used as a basis for building up reduction/science analysis frameworks, so that rather than writing bespoke code for defining the whole pipeline, you could then define a pipeline using a config file (making it easier to test different algorithms and reduction steps).

rosteen:
Cool, thanks for the clarification. I don't have any objections to implementing something like this.

kelle:
I'm having trouble understanding this proposal.  I would be more comfortable starting out at at a lower cost and/or after the ideas have been better flushed out with the rest of the folks involved in `specreduce` and `specutils` development.  Are y'all available to attend the workshop being planned for Malta, post ADASS, as described in #395 ?  

I'd also love to see more discussion of this idea in the #spectroscopy channel.

eteq:

aragilar:
Hi @kelle, sorry I didn't see this message earlier, it got routed into the wrong email address (github unfortunately only allows for one email to be sent notifications). Nuria is on leave currently, so we'll a give a detailed response once she's back later this week.

hamogu:
While I'd live to see more spectroscopic functionality, it is unclear what this proposal aims to do. It suggests a "implementing a generic interface" for specutils. What does that mean? If I want to use an algorithm that's not in specutils (e.g. a FFT or some spectral smoothing, or special filter), then I can call in any other package that I want because the *data structures* are transparent. 
For example, something like this (conceptually):
```Python
tab = astropy.table.Table.read('mytable.fits')
spec = specutils.Spectrum1D(spectral_axis=tab['wave'], flux=tab['flux'])
cont_norm_spec = spec / specutils.fittingfit_generic_continuum(spec)(spec.spectral_axis) 
... other specutils stuff ...
from scipy.signal import hilbert
spec.data = hilbert(spec.data)
```

In that sense, the "generic interface" is already there because I use the generic programming language Python and thus I have the ability to pass the spectrum object (or at least the arrays stored in attributes of that object).

What what this proposal add that I can't achieve this way?

On the other hand, specutils and specreduce themselves would profit from additional astrophysics specific functionality and the group proposing this certainly has the exerience to contribute some of that. I'd be more comfortable with a proposal that adds functionality.

But maybe that's just because I don't understand what the "generic interface" is trying to achieve?

aragilar:
@kelle @eteq Process question: are the proposals "frozen" now, and no changes are to be made (and only comments made), or can we update the proposal along with comments to make the proposal more clear?

aragilar:
@kelle @hamogu I've pushed an update which I hope clarifies where we're coming from. Pulling out the key ideas/aspects as a comment here:

* A significant portion of the project is about getting the redshifting set up cleanly: the FFT cross-correlation works in log-wavelength, whereas most of the astropy/specutils classes are set up to assume wavelength, and a proper solution involves being able to keep track of both without repeated binning and interpolation. It's going to require a fair amount of work to disentangle those assumptions (especially in a well-engineered way, I'm not sure the units framework for example is fully set up for working in log units).
* @hamogu's example is showing why we want to introduce the regularised framework, without modifying the code (keep in mind the median astronomer will not be good at maintaining a proper history in git), there's no way to test different parameter choices. Ignoring the choice of which templates to use, and that due to MARZ being written in JS (so the choice of prep algorithms and their parametrisation is based on what was easy to implement), MARZ has 27 different parameters. Rather than setting them manually in the code, being able to read from a config file (i.e. what most pipeline frameworks do) reduces the amount of random scripts floating around, and allows for improved output formats and provenance tracking.

Currently, at least Simon, Nuria and I are planning to attend ADASS (there will also be other AAO people as well), and so we would attend the Spectroscopy Development Busy Week covered in #395 (I would have attended last year's one had I been aware of it earlier, it was too late to change things like visas when I found out about it).

hamogu:
I like redshift.

pllim:
Sounds very ambitious; one of those high risk, high reward things. Worth a try. Also good to involve more contributors outside of USA and Western Europe.

dhomeier:
I agree it sounds fairly ambitious, maybe not even high risk, but I find it still hard to see where it will lead to.

Since few people are probably familiar with `MARZ`, am I right that this is a feature-matching, redshift _fitting_ code?

And one more question on implementation: would the resampling/binning code build on the existing Resampler functionality in `specutils.manipulation` or add something new?

AnaGabela:
Hi Nuria,

I'm writing on behalf of Astropy's Finance Committee regarding the outcome of your recent Funding Request.

We are sorry to report that your proposal could not be funded during this funding Cycle. We will be closing the FR as a result.

While the community vote was the primary driver of this, it is important to note that this likely does not reflect anything about you as a member of the Astropy Community. Rather, it reflects the specific balance of projects proposed this cycle, the available funds, and the priorities of this cycle’s funding sources relative to others.

We encourage you to read feedback in the thread above or other feedback you might have received via other means, and consider an FR for the next cycle with whatever modifications you think would help. You are also welcome to reach out to the Finance Committee for feedback if you would like.

Ana- on behalf of the Astropy Finance Committee

PR #404, opened 2024-04-12
Do not include travel in Cycle N proposal

pllim:
I vaguely recall a conversation to not include travel in the Cycle N proposal. Does this look about right? Is this the correct document to fix?

cc @kelle @ceb8 et al.


eteq:
We were just talking about this in the finance committee meeting this morning.  The short version is "it might change depending on our funding sources" - the main reason for the change cy4 is differences in Moore vs NASA budgeting, which could change as funding shifts around.  So it might well change again for cy5.

But I agree this is good for at least the current state, so no  reason not to merge it!

PR #434, opened 2024-07-15
remove instructions about master->main transition

eteq:
It's been over 3 years now since this happened, so I don't think there's a need to continue cluttering the README


pllim (approved):
Thanks!

Issue #435, opened 2024-07-19
Proposal to add @svank as a reproject maintainer


astrofrog:
Following the process described [here](https://github.com/astropy/astropy-project/blob/main/policies/adding-new-role-members.md) I would like to propose adding Sam Van Kooten (@svank) as a reproject maintainer. Sam has contributed a significant sub-module (reproject.adaptive) as well as other significant improvements and has the skills and understanding of the code and process to become a maintainer for the coordinated package.

Is it correct that the CoCo will handle the process from this point onwards as described in the policy? Let me know if there is anything I can do to help.

@svank - could you confirm that you would still be happy to be a co-maintainer? (I think Stuart or I chatted with you about this a while back)

astrofrog:
Maintainers of coordinated packages have roles on the team page, and as such I believe they follow the same process as core. The idea of having them be coordinated packages is precisely that their management is overseen by the project.

astrofrog:
The policy at https://github.com/astropy/astropy-project/blob/main/policies/adding-new-role-members.md does not specifically mention the core package, it is a policy for all roles.

pllim:
Okay then. CoCo will meet on 2024-07-22.

pllim:
> I believe they follow the same process as core

If only we could speak for maintenance as well... But that is a battle for another day.

astrofrog:
An additional point to raise with the CoCo is whether the policy could be relaxed so that it is not the CoCo that is responsible for the whole process of filling roles - that is, perhaps any existing role holder could propose someone to hold a role and start the whole 2 week process etc.

pllim:
Re: https://github.com/astropy/astropy-project/issues/435#issuecomment-2239194529

Might go faster if you open a PR with proposed change and CoCo can discuss on Monday.

astrofrog:
@pllim - done in https://github.com/astropy/astropy.github.com/pull/611

pllim:
Oh, I mean proposal to *update the policy*. 😅

astrofrog:
Oh I see, will do

astrofrog:
Hmm actually I'm not really sure about changing it, I'll leave it to the CoCo to think about how they would like to handle this.

pllim:
> I'm not really sure about changing it

What do you mean? Does your request at https://github.com/astropy/astropy-project/issues/435#issuecomment-2239194529 still stand or you have a different request now?

svank:
[...]

PR #436, opened 2024-07-24
Allow any Voting Member to initiate nomination process for adding new role members

dhomeier:
In response to https://github.com/astropy/astropy-project/issues/435#issuecomment-2239194529 this PR is modifying the nomination policy to permit any Voting Member to invite a nominee, open a PR and send the announcement to start the voting period, i.e. the first 5 steps described in the nomination process. It also changes message templates accordingly.

In the present form, this does not specifically mandate any confirmation from the CoCo in this phase, though this could still be added – specifically, if `maintainer_access.md` should still mention "acceptance [of the nomination] by CoCo" this should indeed be confirmed before step 1.

### DIFF
--- a/policies/adding-new-role-members.md
+++ b/policies/adding-new-role-members.md
@@ -13,9 +13,10 @@ which is highlighted in the process below.
 
 The process is:
 
-1. A voting member or maintainer in a related role sends a message to the person nominated
-   (with the Coordination Committee in the cc), making sure they understand the role, 
-   its responsibilities, and confirming that the nominee agrees. 
+1. A voting member or maintainer in a related role, after consulting with other members in
+   related roles, sends a message to the person nominated (with the Coordination Committee
+   in the cc), making sure they understand the role, its responsibilities, and confirming
+   that the nominee agrees.
    ([suggested text](https://github.com/astropy/astropy-project/blob/main/messages/maintainer_access.md)).
 2. The nominee should also be asked (generally but not necessarily in the
    message above) to confirm they agree the Code of Conduct, and that they are
--- a/policies/adding-new-role-members.md
+++ b/policies/adding-new-role-members.md
@@ -13,8 +13,9 @@ which is highlighted in the process below.
 
 The process is:
 
-1. A voting member or maintainer in a related role sends a message to the person nominated (with the CoCo in the cc), making sure they
-   understand the role, its responsibilities, and confirming that the nominee agrees. 
+1. A voting member or maintainer in a related role sends a message to the person nominated
+   (with the Coordination Committee in the cc), making sure they understand the role, 
+   its responsibilities, and confirming that the nominee agrees. 
    ([suggested text](https://github.com/astropy/astropy-project/blob/main/messages/maintainer_access.md)).
 2. The nominee should also be asked (generally but not necessarily in the
    message above) to confirm they agree the Code of Conduct, and that they are
--- a/policies/adding-new-role-members.md
+++ b/policies/adding-new-role-members.md
@@ -13,7 +13,7 @@ which is highlighted in the process below.
 
 The process is:
 
-1. A voting member sends a message to the person nominated, making sure they
+1. A voting member or maintainer in a related role sends a message to the person nominated (with the CoCo in the cc), making sure they
    understand the role, its responsibilities, and confirming that the nominee agrees. 
    ([suggested text](https://github.com/astropy/astropy-project/blob/main/messages/maintainer_access.md)).
 2. The nominee should also be asked (generally but not necessarily in the
--- a/policies/adding-new-role-members.md
+++ b/policies/adding-new-role-members.md
@@ -13,9 +13,8 @@ which is highlighted in the process below.
 
 The process is:
 
-1. A Coordination Committee member sends a message to the person
-   nominated, making sure they understand the role, its responsibilities, and
-   confirming that the nominee agrees. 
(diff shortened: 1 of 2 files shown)


pllim:
From 2024-08-05 CoCo tag-up: Open this to voting member *and* named roles on Team page.

policies/adding-new-role-members.md
```diff
@@ -13,21 +13,23 @@ which is highlighted in the process below.
 
 The process is:
 
-1. A Coordination Committee member sends a message to the person
-   nominated, making sure they understand the role, its responsibilities, and
-   confirming that the nominee agrees. 
+1. A voting member sends a message to the person nominated, making sure they
```
kelle, line 16:
```suggestion
1. A voting member sends a message to the person nominated (with the CoCo in the cc), making sure they
```

kelle (approved):
Looks great! I just made one tweak based on Pey Lian's comment.

policies/adding-new-role-members.md
eteq: including @kelle's suggestion (which I agree with), I'd say
```suggestion
1. A voting member or maintainer in a related role sends a message to the person nominated (with the CoCo in the cc), making sure they
```

eteq (approved):
My small tweak is that I think we should allow maintainers who are *not* voting members to be able to suggest additional maintainers within their domain.

Note in the wording I suggested I just said "maintainer in a related role", which I realize is a bit ambiguous, but that's on purpose: I am just meaning that someone who is not a voting member and also has no connection to the role being proposed should not be encouraged to do this, but in practice that's almost impossible to define cleanly (e.g. does the web site maintainer also count as overlapping with docs?  Or is that not close enough :shrug: ), so we can just say it this  way.

ceb8 (approved):

pllim (approved):
LGTM. Thanks!

policies/adding-new-role-members.md
```diff
@@ -13,21 +13,23 @@ which is highlighted in the process below.
 
 The process is:
 
-1. A Coordination Committee member sends a message to the person
-   nominated, making sure they understand the role, its responsibilities, and
-   confirming that the nominee agrees. 
+1. A voting member or maintainer in a related role sends a message to the person nominated (with the CoCo in the cc), making sure they
+   understand the role, its responsibilities, and confirming that the nominee agrees.
```
dhomeier, line 17:
```suggestion
1. A voting member or maintainer in a related role sends a message to the person nominated
   (with the Coordination Committee in the cc), making sure they understand the role, 
   its responsibilities, and confirming that the nominee agrees. 
```
should spell out here as below since this is the first occurrence

dhomeier:
Sounds like a good solution to me; especially this might include a maintainer for a different coordinated or subpackage that however has a strong dependency on the one needing the new member.

pllim:
2-week comment period has started: https://groups.google.com/g/astropy-dev/c/w9GI11G4H0Q

mhvk (commented):
I like it in general. But I think personally I would tend/prefer to check with others before just sending a message to someone nominating them. Is that something worth mentioning? At some level it is obvious.

mwcraig (approved):

astrofrog (approved):

dhomeier:
> I like it in general. But I think personally I would tend/prefer to check with others before just sending a message to someone nominating them. Is that something worth mentioning? At some level it is obvious.

Hi @mhvk, I just realised that I had probably misread your suggestion, thinking you meant checking with the nominee.
I assume now you mean checking with other members if they support that nomination?
That is also sensible, and perhaps to some degree obvious, but I find it hard to codify in a rule.
Could you confirm if this reasonably phrased?

mhvk (approved):
@dhomeier - your change is exactly what I had in mind - that generally one would check with one or two others before nominating. You want to have a bit of a sense whether the nomination has support beforehand.

pllim:
Given all the approvals and resolved discussion, I merged this. Thanks, all!

pllim:
Should the first paragraph in `adding-new-role-members.md` be updated also?

dhomeier:
Updated to say what?

pllim:
Here is the current wordings:

<hr/>

This document describes the process for adding new people to named roles. This document is *not* about how people are nominated for a role in the first place, but rather the process by which they are added once nominated. The process itself is in the hands of the Coordination Committee according to [APE0](https://github.com/astropy/astropy-APEs/blob/main/APE0.rst).

<hr/>

The "This document is *not* about how people are nominated for a role in the first place" part threw people off. Because they thought only CoCo can make the nomination, which is in conflict with our updated Step 1.

dhomeier:
That first sentence had me confused a bit as well, but not because I read it as “only CoCo may nominate people”, but because it gives the impression the nomination process itself is described in another document. But that document _is_ APE 0, which states
> Anyone can be nominated as a Voting Member by providing evidence of meeting the requirements laid out in the Membership section above. Both self-nominations and nominations by others are allowed.

So nomination has always been open to everyone, nothing about that has changed with this PR.

pllim:
Maybe we should either point to specific section in the APE 0 or repeat the text here, because APE 0 is a lot to read and if people read the wrong parts, they can get confused. What do you think?

pllim:
> Anyone can be nominated as a Voting Member

Wait... when I read the excerpt from https://github.com/astropy/astropy-project/pull/436#issuecomment-2452645047 again, it is about nominating a _voting member_, not a named role on the Team page.

PR #443, opened 2024-10-09
Remove drafts folder

pllim:
Remove drafts folder because it is no longer useful. It currently only has 2 files:

* `code-of-conduct.md` is not needed because any future changes can be made as PR directly to https://github.com/astropy/astropy.github.com/blob/main/code_of_conduct.html
* `interaction-guide.md`, if still used, should be moved somewhere more official like "policies," if we really enforce that during Astropy workshops (lemme know if you want me to move it instead of deleting)


kelle:
Fine with me

pllim:
I opened https://github.com/astropy/astropy-project/issues/450 as follow-up, if necessary. Thanks!

PR #444, opened 2024-10-09
Remove Twitter guidelines

pllim:
Remove Twitter guidelines because we no longer Twit according to @mwcraig .


dhomeier:
For the most part would be applicable to social media engagement in general, so we might want to rewrite it accordingly.
And it would probably be better placed inside `policies`, with a new entry for _Community Engagement Team_.

pllim:
I am not qualified (nor interested) to write a generic social media policy. Can we delete this first and then open a follow-up issue for that? Or do you want to keep this around until someone can get around to doing the generic version?

pllim:
Okay, I opened https://github.com/astropy/astropy-project/issues/449 as follow-up. Merging this one. Thanks!

PR #452, opened 2024-11-04
Clarify named role nomination

pllim:
Resolve the confusion brought up in https://github.com/astropy/astropy-project/pull/436#issuecomment-2452578559

PR #460, opened 2025-02-07
Add policy for adding new role

### LINKED ISSUE #457: Document procedure for adding new roles
I think it would be useful to modify this policy document about adding new people into existing roles to also document the policy for adding new roles (which is VERY similar).

https://github.com/astropy/astropy-project/blob/main/policies/adding-new-role-members.md

@astropy/coordinators

pllim:
Fix #457

Also see https://groups.google.com/g/astropy-dev/c/f3nHkLonlfA

### DIFF
--- a/policies/adding-new-roles.md
+++ b/policies/adding-new-roles.md
@@ -10,6 +10,8 @@ The process is:
    Project and has clear description of associated responsibilities. If this role has sub-roles,
    those must be clarified as well in the same manner. This new role must be declared Unfilled
    for now (separation of concerns). This new role cannot duplicate any existing roles.
+   If the new role touches or modifies the responsibilities of existing roles, those roles,
+   and how they are affected by the proposed new role, shall also be clarified.
 2. The PR author sends out a notice of this PR to astropy-dev mailing list, with cc to the
    Coordination Committee, starting a two-week clock on the feedback period
    [as required by APE0](https://github.com/astropy/astropy-APEs/blob/main/APE0.rst#responsibilities-and-authority).
--- a/policies/adding-new-roles.md
+++ b/policies/adding-new-roles.md
@@ -1,0 +1,23 @@
+This document describes the process for adding new named roles. This
+document is *not* about how people are added to the roles, which is covered under
+[adding-new-role-members.md](https://github.com/astropy/astropy-project/blob/main/policies/adding-new-role-members.md).
+
+The process is:
+
+1. A voting member or maintainer in a related role, after consulting with other members in
+   related roles, opens a PR to [astropy.github.com](https://github.com/astropy/astropy.github.com)
+   to add this new role. The PR must describe clearly why this role is essential to the Astropy
+   Project and has clear description of associated responsibilities. If this role has sub-roles,
+   those must be clarified as well in the same manner. This new role must be declared Unfilled
+   for now (separation of concerns). This new role cannot duplicate any existing roles.
+2. The PR author sends out a notice of this PR to astropy-dev mailing list, with cc to the
+   Coordination Committee, starting a two-week clock on the feedback period
+   [as required by APE0](https://github.com/astropy/astropy-APEs/blob/main/APE0.rst#responsibilities-and-authority).
+3. Once the two week period has elapsed, the Coordination Committee makes a
+   final decision based on any feedback.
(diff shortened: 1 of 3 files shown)


mwcraig (approved):

mhvk:
Looks all good to me!

kelle (approved):
looks good!

aaryapatil (approved):
Looks good. Although, I agree with @dhomeier that we should say Coordination Committee explicitly once this is rebased (refer to [comment](https://github.com/astropy/astropy-project/pull/452#issuecomment-2647767998)).

policies/adding-new-roles.md
```diff
@@ -0,0 +1,23 @@
...
+[adding-new-role-members.md](https://github.com/astropy/astropy-project/blob/main/policies/adding-new-role-members.md).
+
+The process is:
+
+1. A voting member or maintainer in a related role, after consulting with other members in
+   related roles, opens a PR to [astropy.github.com](https://github.com/astropy/astropy.github.com)
+   to add this new role. The PR must describe clearly why this role is essential to the Astropy
+   Project and has clear description of associated responsibilities. If this role has sub-roles,
+   those must be clarified as well in the same manner. This new role must be declared Unfilled
+   for now (separation of concerns). This new role cannot duplicate any existing roles.
```
dhomeier, line 12: This raises  the question how new roles may _overlap_ with existing ones. We probably don't want to simply infringe on those roles' responsibilities, but there can also be legit cases to split the tasks of one roles to several new ones.

Perhaps add something like
> If the new role touches or modifies the responsibilities of existing roles, those roles, and how they are affected by the proposed new role, shall also be named.

May also open the question of a policy for _removing_ roles in general.

dhomeier (changes requested):
Thanks for drafting this! I think all the main points are covered.

In addition to my other comment I suggest to also modify the opening paragraph of `adding-new-role-members.md` to mirror that of the new policy:
```
This document describes the process for adding new named roles. This
document is *not* about how people are added to the roles, which is covered under
[adding-new-role-members.md](https://github.com/astropy/astropy-project/blob/main/policies/adding-new-role-members.md).

```
This document describes the process for adding new people to named roles.
This document is *not* about how new roles are created, which is covered under
[adding-new-roles.md](https://github.com/astropy/astropy-project/blob/main/policies/adding-new-roles.md).
The process itself is overseen by the Coordination Committee.
For filling roles in the Coordination Committee, Ombudsperson, and Voting Members,
please also see the process laid out in
[APE0](https://github.com/astropy/astropy-APEs/blob/main/APE0.rst).
```
Otherwise I think the new file would not be linked from anywhere to be found.
In addition I'd also link it from the "Maintaining (and creating) roles" section in `coco-operating-policies.md`, and perhaps both in fact directly from the README as well, since they now no longer may be initiated by CoCo only.

policies/adding-new-roles.md
pllim: As for removing roles, you mean the assignments to a role, or the role itself, or both? Please clarify at https://github.com/astropy/astropy-project/issues/462 . Thanks!

pllim:
@dhomeier , on your comment about README, I looked at both README under policies and also root level, but I cannot find any relevant text to update. Can you please clarify that request? Thanks!

dhomeier:
> @dhomeier , on your comment about README, I looked at both README under policies and also root level, but I cannot find any relevant text to update. Can you please clarify that request? Thanks!

That's exactly the point; at present those files would only be found when skimming through the repo or the CoCo operating procedures, where a project member not on CoCo may not necessarily look.
My idea was to just add an entry under https://github.com/astropy/astropy-project/tree/main/policies#coordination-committee, but then that is still team-specific...

pllim:
I feel like the README request is out of scope here. This PR should be just about the policy of adding new role. Therefore, I opened a follow-up issue here instead. Hope that is enough for now. Thanks!

* https://github.com/astropy/astropy-project/issues/464

pllim:
(Ooops looks like I put wrong link in my commit message, so we should squash it out on merge.)

dhomeier (approved):
Yes, enough to update the existing links here, thanks!

ceb8 (approved):
LGTM

PR #481, opened 2025-10-01
Add cycle 5

eteq:
This is the announcement for cycle 5 and relevent documents - the SPOC and finance have agreed this is good, with the important caveat that finance needs to get the lastest number for how much we plan to allocate (@kelle is planning to do this in the next few days).  So I'm opening this as a draft until we have that number.


finance/proposal-calls/cycle5/cycle5.md
```diff
@@ -0,0 +1,108 @@
+# Cycle 5 funding requests and ideas
+
+# Changes relative to the previous cycle
+
+The primary change for Cycle 5 (relative to Cycle 4) is that the [Roadmap](https://github.com/astropy/astropy-project/blob/main/roadmap/roadmap.md) will be used to determine funding priorities. In addition, as the maintainers of the roadmap, the Strategic Planning and Organizing Committee (SPOC) will participate in funding request review, solicitation, and selection. Also, due to funding uncertainties, the period of performance and budgets should be just one year, as opposed to multiple years.
+
+# Funding goals
+Funding requests (FR) should address items on the Astropy Project’s Roadmap. In particular, funding will be prioritized for “red” and “orange” roadmap items, which have been identified by the community as being high priority.
```
astrofrog, line 8:
```suggestion
Funding requests (FR) which address one or more items on the Astropy Project’s Roadmap will be prioritized, with proposals that include work towards “red” and “orange” roadmap items being prioritized the most, since these have been identified by the community as being the most critical.
```

Or could say 'should address one or more' if we want to require they do

finance/proposal-calls/cycle5/cycle5.md
```diff
@@ -0,0 +1,108 @@
...
+* Nov 7, 2025: Finalized FRs are due, two week Voting period begins
+* Nov 24, 2025: The SPOC and Finance Committee will work together to select funding requests and allocate budgets. 
+* Dec 8, 2025: Announcement of selected FRs. Drafting contracts begins.
+* Jan 1, 2026: Work commences
+
+Note that these dates are nominal, and unplanned events like a government shutdown could extend some of the dates/deadlines.
+
+We aim to achieve broad consensus before the vote. Requesters are encouraged to combine ideas and teams, modify, or withdraw FRs during the discussion period, as appropriate. Discussion is open to anyone; it is not restricted to voting members. 
+
+Voting members of the Astropy Project vote on the FRs using thumbs up/down on a dedicated comment. Every voting member of the Astropy Project has equal voting rights.
```
astrofrog, line 45: I think I asked this in the past, but are we going to actually check the votes against the voting member list? If so maybe could then also say "Non-voting members are free to vote on FRs, but these votes will not be included in the count to determine funding allocations"

astrofrog (commented):
@eteq - here are my suggestions - feel free to word these differently though

PR #483, opened 2025-10-29
Cycle 5: Quantities with Array API support, Improved Support for Masks and Uncertainties

mhvk:
I request continued partial buy-out from my professorship at UofT to be able
to work one day a week on projects that are too large for the time I can
otherwise commit for astropy.  Specifically, I propose,

- Facilitate Quantity becoming a container class that can handle not just
  ndarray but any type of array, i.e., also dask, jax, etc.
- Ensure Quantity is fully compliant with the new Quantity API being developped.
- Extend the same machinery to Masked and Distribution so that all main astropy
  classes can use arbitrary array classes.
- Also extend the machinery to the internal arrays used by Time.
- Speed up unit conversion and thus all of astropy by smarter conversion functions
  and caching.
- Finish my implementation of a Variable class that tracks uncertainties and
  their correlations analytically (based on the uncertainties package).

### DIFF
--- a/finance/proposal-calls/cycle5/mhvk-units-masked-uncertainties.md
+++ b/finance/proposal-calls/cycle5/mhvk-units-masked-uncertainties.md
@@ -18,8 +18,8 @@ otherwise commit for astropy.  Specifically, I propose,
 - Extend the same machinery to Masked and Distribution so that all main astropy
   classes can use arbitrary array classes.
 - Also extend the machinery to the internal arrays used by Time.
-- Speed up unit conversion and thus all of astropy by smarter conversion functions
-  and caching.
+- Speed up unit conversion and thus all of astropy by smarter conversion
+  functions and caching.
 - Finish my implementation of a Variable class that tracks uncertainties and
   their correlations analytically (based on the uncertainties package).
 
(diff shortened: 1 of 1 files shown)


finance/proposal-calls/cycle5/mhvk-units-masked-uncertainties.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Period of Performance
+
+Ideally, I would be covered until June 2027, which is the end of an academic
+year.
+
+I note that the funding provides me with teaching relief, which is for one
+semester of an academic year (where thus more than 1 day/week is spent on
+astropy, while less time is spent when I teach).  So far, the relevant
+semesters have been July-December 2024 and January to June 2026.  It may be
```
kelle: Do you mean 2025 here? 
```suggestion
semesters have been July-December 2024 and January to June 2025.  It may be
```

kelle:
Please include the remaining approved $36,000 from your Cycle 4 proposal in this one.  I believe that would bring the total request to $90,000 for work done in FY25 and FY26.

kelle:
I think this request would benefit from providing a budget range. For example, the teaching release in the Spring is already confirmed, so a minimum budget of $36,000 might makes sense. (That was also already approved in Cycle 4.)

eteq:
We are pleased to let you know that this proposal will be funded at the level of $36,000. The Astropy community supports this work and the SPOC finds it to be well-aligned with the roadmap. While we would ideally fund this work in full, given over-subscription and budget limitations, this proposal will be funded at less than the maximum budget request. With the amount allocated, we ask that you focus your funded work to address the roadmap items listed in the request. If the amount allocated is insufficient to make any progress on the roadmap items, please let us know.

Additionally, assuming a successful and timely budget process with NASA, we are anticipating more funds to be available at some point in 2026. Once this is final, the SPOC and Finance Committee will review project progress and spending to date and evaluate the possibility of modifying budgets mid year. At that time we may be able to allocate more funding to this project.

We will follow-up with more details on how to start working on this in the near future. In addition, individualized feedback will be forthcoming.

PR #382, opened 2024-02-29
Cycle 4: Astroquery & other maintenance (Ginsburg)

keflavich:
This proposal covers maintenance effort by me on astroquery, astropy.convolution, reproject, regions, and affiliated packages.

PR #484, opened 2025-10-29
Cycle 5: Astroquery maintenance (Ginsburg)

keflavich:
This is the same request I proposed last year and had approved as a 3-year project, but per the guidance on the astropy slack, I'm re-submitting fresh for 1-year.

See also https://github.com/astropy/astropy-project/pull/382.


eteq:
@keflavich - while this does certainly relate to some roadmap items, it looks like you didn't explicitly list any - you may want to add that (I gather this was just adapted from last cycle, but we updated the template this cycle to add a prompt to state which part of the roadmap the FRs connect to)

PR #489, opened 2025-10-30
Cycle 5: Continuation of the Matrix <> Slack Bridge

Cadair:


pllim:
> 10% of active users

So we're talking about 20 people?

Issue #411, opened 2024-05-02
Cycle 4 Funding: Array API and astrowidgets (Craig)


jdswinbank:
Issue to track progress of Cycle 4 funding request which was approved in https://github.com/astropy/astropy-project/pull/393 .

- [Project Description](https://github.com/astropy/astropy-project/blob/main/finance/proposal-calls/cycle4/mwcraig-cycle4-ccdproc-astrowidgets.md)
- Awarded Amount: $11,250 (US)
- Period of Performance: 2024-05-01 through 2025-04-30
- Finance Committee Contact: @jdswinbank 
- COTR: @pllim

AnaGabela:
Hey Matt

Let's get you started on getting a new contract with NumFocus because I think that's gonna take them a little while. Here are the instructions:

Go to https://numfocus.org/rocket
Click on Independent Contractor Registration
Form fields
Project: Astropy
Grant: NASA
Full Name
Email address
Street Address
Street Address
City
State
Postal Code
Country
Start Date: May 1, 2024-Feb 28, 2025
Description: Specreduce and spectroscopy development
Project Contact name: Astropy Finance Committee
Project Contact email: [finance@astropy.org](mailto:finance@astropy.org)
Payment method: ACH, Wire Transfer, PayPal
Bank Information
Comments about payment
Upload W9 or W8
Please let me know that you have submitted the form so we can keep track of it.

mwcraig:
Hi @AnaGabela -- I just filled out the contractor form.

pllim:
Just so we're clear what I should be tracking, `Awarded Amount: $11,250 (US)` means Astropy is only funding "`ccdproc` and `astropy.nddata` work" section of https://github.com/astropy/astropy-project/blob/main/finance/proposal-calls/cycle4/mwcraig-cycle4-ccdproc-astrowidgets.md ?

AnaGabela:
Yes I only see the Awarded Amount: $11,250 (US)

pllim:
Ah, I see https://github.com/astropy/astropy-project/pull/393#issuecomment-2090515053 now but unclear from the comment what exactly is being funded for a year. @jdswinbank , can you please clarify? Thanks!

jdswinbank:
> ... Astropy is only funding "ccdproc and astropy.nddata work" section ...

This is correct.

pllim:
> I think it would be appropriate for mwcraig to choose

Should this be decided upfront so we can set the expectations early on?

[...]

PR #490, opened 2025-10-30
Cycle 5: Array API in `ccdproc` and `NDData` and work on astrowidgets

mwcraig:


kelle:
Are the funds requested in addition to the amount remaining in your Cycle 4 project? or an updated estimate for the cost of the work during 2026?

mwcraig:
One note about the relationship  of this proposal to #483. While I have some changes to do in `astropy.nddata` to get `ccdproc` as close as possible to implementing the Python Array API making it fully compatible will be far easier if `Quantity` supports the Array API. I would not go so far as to say that the Array API-related work in this proposal cannot proceed without #483 I think it is fair to say that the work will be much easier once `Quantity` has been revised.

(pinging @Mhvk just so he is aware of the connection)

AnaGabela:

mwcraig:
I want to clarify something that I maybe have not been explicit about: most of the work at the minimum budget level for this proposal has already been done as part of this [approved proposal for Cycle 4](https://github.com/astropy/astropy-project/issues/411). I was asked to hold off on invoicing for that work until 2026 because of the cashflow issue the project had because of the delay in obtaining the new ROSEZS funding.

So on the one hand I guess I would meet all of the deliverables I outlined in the minimum budget scenario here even if the budget was reduced 100%, but on the other hand I'd feel a little bitter about that....

PR #494, opened 2025-10-31
Cycle 5: Improving `astropy` units parsing

eerovaher:
I am submitting a finance request for work on `astropy` units parsing, in particular parsing function units and custom units. The latter is important for reading existing data files that might be using non-standard units.

### DIFF
--- a/finance/proposal-calls/cycle5/eerovaher-unit-parsing.md
+++ b/finance/proposal-calls/cycle5/eerovaher-unit-parsing.md
@@ -1,0 +1,68 @@
+### Title
+
+Improving `astropy` unit parsing
+
+### Project Team
+
+Eero Vaher
+
+### Project Description / Scope of Work
+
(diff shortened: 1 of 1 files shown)


nstarman:
Yes, it's definitely challenging, some kind of deprecation would be necessary.
Pinging @mhvk for thoughts...

dhomeier:
Agreed that `UnrecognizedUnit` is not helpful at all here, but custom unit does not seem so far off – it is not an allowed CDS unit, and there is no obvious general way how this should be represented. Your example has some potential for confusion, as I would intuitively expect `"10-3m"` to become milimetres, but would you interpret `"2-3m"` as something like "octimetres"? Realising that `"10-4m"` is permitted in CDS as well, you may indeed say let `parse_strict!="silent"` pass this as `Unit("0.125 m")`, but this would already get you in trouble for cases like `"7-1m"` (represent this as a fraction rather? how many significant digits...?)

PR #500, opened 2025-11-01
Cycle5: Improving astropy vo and votable capabilities

bsipocz:
This is proposal is very much a draft and will benefit from discussions at the upcoming IVOA interop.

Also, no language tuning was made, sending through  the AI machinery could improve the fluff.

@tomdonaldson @andamian -- I mentioned 3 of us in the text as it's clear we cannot take on contracts so at best we can coordinate this project (by definition, if we do it well, it should not add much burden on the existing maintainers team).

Also, cc @ManonMarchand and @msdemlei, I hold on hope that you could take on the contract if it's selected, but either way, let's discuss it in Görlitz


bsipocz:
thank you so much @ManonMarchand!

pllim:
I don't want to accidentally both the beautiful list above, so just FYI that if you turn them into bullet points, GH will render also the issue titles and state (close/open) for easier future reading.

PR #503, opened 2025-11-01
Cycle 5:Accessible stellar photometry

mwcraig:


AnaGabela:
Hi,
Can you let us know what is the hourly rate you will be charging? We will need that for the paperwork.

Also can you clarify who will be the payee? Do you have a name, other than you, who will be paid for this project?

Thanks!
A

mwcraig:
Here are the things I think could be done with partial funding of the minimum (we would get there by reducing the deliverables though):

1. With a 25% reduction from the minimum budget (i.e. funded $3,375), I would offer pay for the outside testers/users at the hgiest rate I could. The postage stamp feature would be dropped.
2. At 50% reduction from the minimum budget (i.e. funded $2,250), I would expect there will be fewer testers because of the reduced rate of pay. Workshop and tutorials would not be developed, in favor of simply building out the current documentation.
3. If the minimum budget is reduced by 100% (i.e. no money I guess), I anticipate adding an ubercal-style calibration no matter what. Much of the work is done, though some technical details need to get sorted out and I need to see how well it scales. I'd still probably invite folks to kick the tires on the software, no idea what sort of uptake I'd get.

PR #504, opened 2025-11-03
Cycle 5: Astropy UX User Interviews

eteq:
This is 1 business day late for the draft deadline.  I had some technical and logistical issues on Friday that prevented me from submitting it then, but anyone should feel free to weight this late-submission appropriately to their own thinking when considering whether they support the proposal.

### DIFF
--- a/finance/proposal-calls/cycle5/tollerud-user-interviews.md
+++ b/finance/proposal-calls/cycle5/tollerud-user-interviews.md
@@ -1,4 +1,5 @@
 ### Title
+Astropy UX User Interviews
 
 ### Project Team
 Erik Tollerud will lead the project, Jenn Kotler will do most of the interview development and results summary (and is who all of the funding would go to).  If she cannot be funded for logistical/financial reasons, another UX specialist at an astronomy institution, university, or archive will be hired for this project.


kelle:
I would be interested in know more about the proposed contractor, Jenn Kotler.  Maybe link or include their resume?

AnaGabela:

kbwestfall:
Just to offer a contrasting opinion to @mhvk : As part of an NSF award, the pypeit team was required to participate in the I-Corps for POSE training, which requires each team to conduct 60 "ecosystem discovery" interviews over a period of 4 weeks (@eteq was one of the people we interviewed!).  Personally, I found these interviews really useful, but they were a ton of work.  There *can* be value in these interviews, and I think it helped us set development priorities.  But here are some things to consider: (1) I would argue you need more than 10 interviews.  I think 60 was too many for us, but I think 10 would have been too few.  (2) This is discussed broadly in the proposal, but *who* you interview will be very important, particularly if you only do 10.  Instead of a broad target (astronomers with a "reasonable geographic and demographic spread"), I'd recommend identifying specific groups you think might have been under-represented in the user survey.  (3) I found it easy to fall into the trap of listening for what I wanted or expected to hear during the interviews.  I don't have any specific advice on how to avoid this, but I would find it useful to keep this in mind when developing the interview script.

Having said all that, pypeit and Astropy are rather different projects, so what's true for one won't necessarily be true for the other...

mhvk:
@kbwestfall - thanks, that's interesting to hear. I do think the projects are very different, astropy being really more like numpy, scipy and matplotlib. So much so in fact that it is very hard to get people to think of it as something they could contribute to rather than just use it. I do share your question about how the 10 would be selected...

tepickering:
> @kbwestfall , yes I would be interested to see. This is because I understand that pypeit has an visual component, so if the takeaways are all about visualization, I suspect it wouldn't overlap much with "Astropy UX".

UX does not necessarily mean GUI. it's overall "user experience", how they interact with the system as a whole.

kbwestfall:
> @kbwestfall , yes I would be interested to see. This is because I understand that pypeit has an visual component, so if the takeaways are all about visualization, I suspect it wouldn't overlap much with "Astropy UX".

Just posted a couple of slides in Slack.

hamogu:
The comment I heard was "Astropy is constantly breaking backwards compatibility, thus we should not build on top of it". This was in a zoom meeting with about a dozen people on a committee tasked to shape the future of NASA's high-energy software and data archives at HEASARC. Since this was not an astropy meeting, I did not deem is appropriate to interject and say "Can you please specify that in detail and give examples of things that you relied on that broke?", so I don't really have much more information. All I can say is that several people on the call nodded approvingly to that comment.

It's not an official recommendation to "not use astropy". To the contrary, the official recommendations will likely be something along the lines of "intensify collaboration and interoperability with other open source projects in the astronomy community". However, I think a comment like that does show a sentiment that is shared by a significant (though hopefully minority) fraction of the community.

Because I don't have more details, I can't say what that specific person had in mind. However, I can give some general thoughts: In < 15 years, astropy went for version 0.4 to 7.2 and if you have a astropy 0.4 (or even 1.0 or 2.0) script, it's almost impossible that that still works - back than astropy was on python 2.x. Even Python code that relies on astropy that's only 10 years old, is unlikely to work to day - sure, most individual functions are still similar, but if you write something more complex, e.g. a data reduction pipeline for some specific paper you are likely to have hundreds or thousands of LOC and somewhere in this code, there will some something that doesn't work any longer, sometimes silently, for example, we changed the default cosmology to pre-publication Plank 2018 in 4.0, and back than, most people probably didn't explicitly set or log their cosmology, but just used the default, in 4.0 (just to stick with on example), we also changed the galactocentric frame, changed the entire modeling API, and tables don't upgrade to masked tables any longer just because one column has a mask.
And 4.0 wasn't even 10 years ago!

In contrast, HEASARC's ftools (library and command line tools to manipulate fits files, often used as building block for data reduction pipelines for X-ray missions such as NICER or IXPE), is essentially unchanged compared to 1990 or so (with some functions added), XSPEC (the default X-ray fitting program) went for version 11 to 12 (today 12.15.1) in the last 15 years - compared that to astropy. The lifetime for software is just long, for example Chandra's CIAO was designed in the 1990's pre-launch. Now, Chandra's software is one example of a software that was redone several times to keep up with general developments (for example, switching form S-Lang to Python for scripting about 15 years ago), but each switch costs years of development time that is lost for other improvements. 

So what we think of improvements and "long" depreciation times are just the blink of an eye compared to the lifetime of large missions. Chandra, XMM, HST, all have embraced Python to some degree, but even HST, probably the most Pythonic of the large observatories) took about two decades to do it (helping to invent numpy and matplotlib in the process). So, if   manager that's planning for a mid-size observatory (i.e. one that will produce data for the public and fly for a few years, say wit ha launch data in 2030 or so) today would consider stable, they would probably a software that doesn't change noticeably between now and mission end in 2035. In other words, they would not care if we do and LTS for 6 months or 1 years, they would want an LTS for 10 years.
On the ground, many mid-size observatories (that, unlike ESO or Gemini, don't have full professional software teams) still use IRAF, abbot two decades after the official development of that software ended.

We just can't offer that in Python. Python itself is changing much faster than that. Maybe the first usable Python 3 was 3.3 from 2012. And home many packages today are backwards compatible to Python 3?

PR #506, opened 2025-11-11
Update CoCo transition checklist on transferring Gsuite ownership

dhomeier:
Adding instructions on granting Gsuite ownership rights and astropy.org accounts.

Hope this makes sense where it is placed now between ownerships of the Google Shared Drive and the various mailing lists (I really don't understand any of the details how these are handled by Google).


kelle:
No, giving someone access to manage the groups step needs to be done by hand. No magic there.

ceb8 (approved):

PR #508, opened 2025-11-13
Cycle 5: Reliable performance benchmarking for the Astropy project

astrofrog:
This proposal was solicited by the Strategic Planning Committee as it was identified that no existing proposal addressed the roadmap item about performance benchmark reporting (hence why this was opened after the deadline).

Note that this is separate from the NumFOCUS approach which @pllim mentioned in https://groups.google.com/g/astropy-dev/c/Ns2jj7qtW-s - the approach in the current proposal has a lower monthly cost and includes developer time to make it happen.

### DIFF
--- a/finance/proposal-calls/cycle5/aperio_benchmarking.md
+++ b/finance/proposal-calls/cycle5/aperio_benchmarking.md
@@ -1,0 +1,42 @@
+### Reliable performance benchmarking for the Astropy project
+
+### Project team
+
+This proposal is to fund the following people at [Aperio Software](https://aperiosoftware.com):
+* Thomas Robitaille
+* Stuart Mumford
+This could also fund other individuals working at Aperio Software if required.
+
+### Project Description / Scope of Work
+
+The goal of this project is to set up a dedicated physical server in a way that can be used to run the astropy benchmarks.
+
+#### Roadmap Items
+
+This project will address the following roadmap item:
+
+- :large_orange_diamond: Maintain and improve robust performance benchmark reporting.
+
+#### Project / Work / Deliverables
+
+The main tasks in this project are:
+
+* Rent a commercially available dedicated physical server
+* Set up the server as a GitHub self-hosted runner (taking care to bind the runner to a specific core if possible to ensure stable timings)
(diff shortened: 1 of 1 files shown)


astrofrog:
I think the best way to guard against this is to have the astropy project pay directly for the server once we have identified which one to get and for us to openly document the server set up or at least if not completely open, have it in a repo that at the very least the CoCo have access to. Then anyone else can take over maintenance of the server. We plan to have the server side set up be as simple as possible, with all the important config being in eg the core package repo.

astrofrog:
Part of the reason that the benchmarks have not been as useful as they could be is because it's not been possible to really run them reliably as part of the CI - yes we do have it running on astropy core in CI as opt-in (as it requires a label), but we have a low bar for looking at regressions because of the lack of dedicated/stable hardware. If we could run the benchmarks in a more stable environment, we could be more precise in terms of timings and look for smaller changes. We could also in principle run the benchmarks on all PRs as sometimes it is unexpected PRs that will introduce regressions.

Cadair:
I also use codspeed for dkist and it's good.

I would say the main drawbacks of it are:

* The free tier can only use their instrumented benchmarking which can't benchmark syscalls so things like io.
* It can't benchmark memory usage.

I will clarify that the main objective of this FR is to provide dedicated hardware for stable benchmarking as a github actions runner. This means we could use it for asv, pytest-benchmark or codspeed (I think).

eteq:
We are pleased to let you know that this proposal will be funded at the level of $6600. The Astropy community supports this work and the SPOC finds it to be well-aligned with the roadmap. While we would ideally fund this work in full, given over-subscription and budget limitations, this proposal will be funded at less than the maximum budget request. With the amount allocated, we ask that you focus your funded work to address the roadmap items listed in the request. If the amount allocated is insufficient to make any progress on the roadmap items, please let us know.

Additionally, assuming a successful and timely budget process with NASA, we are anticipating more funds to be available at some point in 2026. Once this is final, the SPOC and Finance Committee will review project progress and spending to date and evaluate the possibility of modifying budgets mid year. At that time we may be able to allocate more funding to this project.

We will follow-up with more details on how to start working on this in the near future. In addition, individualized feedback will be forthcoming.

PR #530, opened 2026-02-09
AI Policy

### LINKED ISSUE #509: Policy against AI-generated pull requests
With the rise of popularity of using AI/LLM in coding, we have since a rise of pull requests submitted without true understanding of the code nor the problem. Some of them may look legit at first glance, only revealing flaws when one dive into the code/logic during the review process. To pick apart such flaws sometimes require domain expertise.  On the other hand, some may overwhelm the system by opening a plethora of pull requests within a very short span of time; even if these are obviously dead on arrival, they still take time to triage. All in all, they introduce extra maintenance burden to maintainers who are already stretched thin.

Therefore, Astropy should join some other projects in discouraging such AI submissions. Examples we could adopt from, include but not limited to:

* https://devguide.python.org/getting-started/generative-ai/ 
* https://github.com/scikit-learn/scikit-learn/blob/main/CODE_OF_CONDUCT.md#low-quality-and-ai-generated-contributions-policy 
* https://github.com/sunpy/sunpy?tab=readme-ov-file#usage-of-generative-ai 
* https://github.com/zulip/zulip/blob/main/CONTRIBUTING.md#ai-use-policy-and-guidelines 
* https://llvm.org/docs//AIToolPolicy.html

ceb8:
Adding an AI policy document to Astropy's policies.

This policy was drafted by the CoCo, making use the discussion to close https://github.com/astropy/astropy-project/issues/509

The comment period for this policy document is open now (9 February), and will run through 23 February.

### DIFF
--- a/policies/ai-policy.md
+++ b/policies/ai-policy.md
@@ -1,0 +1,11 @@
+Usage of Generative AI
+
+Astropy maintainers welcome contributions, including those developed with the assistance of generative AI tools. However, the use of such tools does not change our expectations for contributor responsibility, engagement, or code quality. The following apply to all contributions.
+
+Human ownership and accountability
+Contributors are responsible for all submitted content, regardless of whether generative AI tools were used.
+Contributors must be able to explain all changes during review.
+
+Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback.
+
+Consistency with Astropy conventions and existing style. In the context of Generative AI, this particularly means conveying information (comments, documentation, etc) in a style the Astropy user community will understand.
--- a/policies/ai-policy.md
+++ b/policies/ai-policy.md
@@ -4,6 +4,10 @@ Astropy maintainers welcome contributions, including those developed with the as
 
 1. Human ownership and accountability
     1. Contributors are responsible for all submitted content, regardless of whether generative AI tools were used.
-    1. Contributors must be able to explain all changes during review.
-2. Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback.
-3. Consistency with Astropy conventions and existing style. In the context of Generative AI, this particularly means conveying information (comments, documentation, etc) in a style the Astropy user community will understand.
+    2. Contributors must be able to explain all changes during review.
+2. Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback themselves. Copying and pasting replies to / from a generative AI tool does not count as engaging with the reviewer; contributors are expected to understand the changes they are proposing, not act as message courier for a third party. 
+3. Consistency with Astropy conventions and existing style. In the context of Generative AI, this particularly means conveying information (comments, documentation, etc) in a style the Astropy user community will understand, and that is appropriate and to the point.
+
+> [!NOTE]
+> This policy is in part inspired by the scikit-learn ["Automated Contributions Policy"](https://scikit-learn.org/stable/developers/contributing.html#automated-contributions-policy), which has additional details that may be useful in interpreting this policy.
+
--- a/policies/ai-policy.md
+++ b/policies/ai-policy.md
@@ -1,11 +1,9 @@
-Usage of Generative AI
+# Usage of Generative AI
 
 Astropy maintainers welcome contributions, including those developed with the assistance of generative AI tools. However, the use of such tools does not change our expectations for contributor responsibility, engagement, or code quality. The following apply to all contributions.
 
-Human ownership and accountability
-Contributors are responsible for all submitted content, regardless of whether generative AI tools were used.
-Contributors must be able to explain all changes during review.
-
-Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback.
-
-Consistency with Astropy conventions and existing style. In the context of Generative AI, this particularly means conveying information (comments, documentation, etc) in a style the Astropy user community will understand.
+1. Human ownership and accountability
+    1. Contributors are responsible for all submitted content, regardless of whether generative AI tools were used.
+    1. Contributors must be able to explain all changes during review.
+2. Authentic engagement. The pull request process is collaborative and iterative. Contributors are expected to actively engage with reviewer feedback.
+3. Consistency with Astropy conventions and existing style. In the context of Generative AI, this particularly means conveying information (comments, documentation, etc) in a style the Astropy user community will understand.
--- a/policies/ai-policy.md
+++ b/policies/ai-policy.md
@@ -5,8 +5,9 @@ Astropy maintainers welcome contributions, including those developed with the as
 1. Human ownership and accountability
     1. Contributors are responsible for all submitted content, regardless of whether generative AI tools were used.
     2. Contributors must be able to explain all changes during review.
+    3. Contributers are responsible for ensuring all submitted code complies with Astropy's licensing, regardless of whether generative AI tools were used.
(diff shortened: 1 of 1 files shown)


policies/ai-policy.md
```diff
@@ -0,0 +1,9 @@
+# Usage of Generative AI
+
+Astropy maintainers welcome contributions, including those developed with the assistance of generative AI tools. However, the use of such tools does not change our expectations for contributor responsibility, engagement, or code quality. The following apply to all contributions.
+
+1. Human ownership and accountability
+    1. Contributors are responsible for all submitted content, regardless of whether generative AI tools were used.
```
mhvk: Should this include something like ", including that the code does not break copyright" - AI is known to copy verbatim code that falls under the GPL.

mhvk (commented):
Thanks for writing this! Two small comments.

In the end, I'm not sure this addresses my main worry, which is poorly thought-through contributions with minimal improvements by people who have no intent to contribute further, making review essentially a waste of time. But really that is independent of AI, so out of scope here; it is more a question of community building (and how to ensure reviewing does not become a chore).

astrofrog (approved):
I’m happy with this as a starting point and we can always review in future if it is not sufficient.

policies/ai-policy.md
neutrinoceros: Indeed, but obvious cases are frequent enough that it would still be useful, IMO

Cadair:
Also the PR template?

policies/ai-policy.md
```diff
@@ -0,0 +1,9 @@
...
+    1. Contributors must be able to explain all changes during review.
```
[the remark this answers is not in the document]
zacharyburnett: all the same, not everyone reads markdown in codebases rendered (I know I often read `CONTRIBUTING.md` in my editor after cloning locally)

Cadair (changes requested):
I'm going to explicitly say here I don't think this goes far enough. I'm not completely sure I know how far I think we should go at this point but this isn't it.

Scikit-learn and llvms policys are a lot closer to where I think we need to be.

bsipocz:
> I'm going to explicitly say here I don't think this goes far enough. I'm not completely sure I know how far I think we should go at this point but this isn't it.
>
> Scikit-learn and llvms policys are a lot closer to where I think we need to be.

Full agree with Stuart on all of this! AI is not our friend when it's about library development. Do hack away with it for a website deployment, but here we talk about soaking up a lot of scarce maintainer resources and goodwill on onboarding new "people".

zacharyburnett:
Perhaps blocking users from collaborating with the organization should be a more readily-used option if they waste maintainer time with slop PRs and comments (as in, it's clear that they are not engaging with the maintainers themselves). In my (limited) experience with users who submit slop, they usually aren't sources of quality code contributions in the first place, even without using genAI, so indiscriminately blocking them doesn't really seem like a big loss for external collaboration

However, I'm not an active contributor to `astropy` and don't really have a feel for this community at large, so take my opinion with that in mind

mhvk:
> Perhaps blocking users from collaborating with the organization should be a more readily-used option if they waste maintainer time with slop PRs and comments

The problem is that this effectively forces the maintainers in a kind of policing role (this is a point made by David Hogg in his arXiv white paper on astronomy more generally). 

As a maintainer, if every PR from a new contributor has a substantial likelihood (>10%?) of being AI slop, I'll end up not looking at new-contributor PRs anymore (actually happening already...). I'd still prefer a "web of trust" system were we make people introduce themselves -- then, it just happens once, at first entry. Obviously, that poses an extra hurdle, but I think it can also be a way to more immediately make a connection, leading to more friendly review. (And, yes, there will still be bad actors, but surely fewer; lying is different from submitting something that an AI tells you is a good solution.)

eteq:
@Cadair 

> I think we can have documentation with strict guidelines on what we deem to be acceptable and still only enforce them when we want to. This is the case for many of the technical guidelines we have as well.

Ah I see your point here - totally agree this is similar to e.g. style guidance where sometimes exceptions are ok if they are for good reasons and its left to maintainers to make the judgement call.

> For me the main thing this is lacking is some strong words on not pasting the output of a LLM into the comments / OP of a PR. This is the the most egregious thing imo.

I really like this idea!  Hopefully @ceb8 can add something to this effect. And combined with the above point maybe this helps with some of the concerns @bsipocz raised? (it doesn't solve them all but at least starts us down a path?)


@mhvk

I think the idea of web-of-trust or similar is a good one, except I'm not as sure we can implement it usefully in a context like Astropy where there are quite different communities that have overlapping but different perspectives that are both needed (i.e. the engineering/developer/technical community and the science community). That is, even if we made a successful web-of-trust model for, say the astronomy research community, it might exclude technical experts who aren't known in the research community but are still valuable contributor/maintainers.  It also has the potential to further exacerbate some of the existing issues where e.g. people in the Global South are underrepresented because they aren't plugged into the right communties. I'm sure there are solutions to this but they are not something that I think we can do *right now*. (which I guess  is maybe your point of why we want you want to focus on that instead of this policy?)

mhvk:
@eteq - agreed that a web of trust needs some thought, which is why I suggested with starting with something like having a welcome message that asks new contributors to introduce themselves, to try to get going with a human connection asap.

To me, the details of the AI policy are less important, since I do not expect them to be read, i.e., the burden is still on the maintainer to be policing it. The only help is that at least there is something to point to.

Aside: the discussion in numpy is (somewhat) going in the direction of adopting the scipy policy (or perhaps a version of llvm); see https://scipy.github.io/devdocs/dev/conduct/ai_policy.html

Personally, I like the idea of a large part of the scientific python ecosystem having the same policy (and making things easier for ourselves by adopting instead of rewriting).

bsipocz (commented):
I don't think it's just about the user community, but about the whole community. And also it's not about our understanding the comments/documentations but to have a preferred style we use/adhere to.

mhvk (commented):
@ceb8 - thanks for the iteration! I think with @bsipocz's addition, this is pretty good.

That said, in the (not yet completed) discussion about a numpy variant of this, we are tending to simply adopt an AI policy, I do wonder whether that is not the better path for astropy too. It has the advantage of moving towards a state where the AI policies across the scientific python ecosystem align.

In particular, I find the [scipy AI policy](https://scipy.github.io/devdocs/dev/conduct/ai_policy.html) a rather nice one. It is short and concise, yet well organized, and it has a few things that the current text does not have (at least not explicitly):
1. It asks to disclose the use of AI.
2. It discusses "AI slop" and explicitly asks not to waste maintainer time.
3. It discussed copyright explicitly.
4. It explicitly forbids using AI agents.

Anyway, I am OK with going with this policy, but prefer simply adopting the scipy one.

aaryapatil (approved):
Community input seems to be addressed properly. Looks good on my end.

eteq (approved):

ceb8:
The CoCo has decided to proceed with the current document rather than adopting that of Scipy, because we wanted to start with something simpler and then reassess once we see how it functions. To this end we will discuss the policy, and how we do or do not want to change it, including whether we want to adopt another project's policy, at the coordination meeting.

Thanks so much everyone who gave feedback!

pllim:
Thanks, all!

Also, do you notice all the non-astropy repos linking to this? 👀

zacharyburnett:
there's this in-development github action that purports to compare pull requests against a set of metrics (number of emojis, description length, number of code changes, etc.) and give reviewers a score of how much the PR smells like AI; that probably won't catch _all_ of the AI slop PRs, but it might make it easier for / reduce the cognitive load on reviewers and maintainers when looking at PRs

https://github.com/peakoss/anti-slop

